PR #24, opened 2020-02-01
Launch MySQL Server before entering credentials

MichelBaie:
It gives an error if it's not started !


VilhelmPrytz:
Hmm, what OS are you experiencing this issue on? I believe the MySQL daemon is not named `mysqld` on all systems.

PR #26, opened 2020-02-17
Add support for Zorin OS (panel)

Iomega0318:
I was experimenting with Zorin as that's all I had laying around on a USB Stick at the time, as the OS is built on top of Ubuntu 18 there isn't much difference in the OS other than the overhauled UI. So I changed your script to include it locally. Just suggesting the change in case anyone else uses this. I was able to install and use it with those changes.

### DIFF
--- a/install-panel.sh
+++ b/install-panel.sh
@@ -124,6 +124,13 @@ function check_os_comp {
     else
       SUPPORTED=false
     fi
+  elif [ "$OS" == "zorin" ]; then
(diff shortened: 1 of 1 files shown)


VilhelmPrytz:
Hmm, I'm not too sure why you'd want to install pterodactyl on a desktop distribution, but hey, why not? :smile:

As far as I can see, this looks all good, I'll try this tomorrow when I have time (and merge it then)!

PR #27, opened 2020-02-17
Add support for Zorin OS (daemon)

Iomega0318:
Updates to go along with the Zorin updates for the panel.

### DIFF
--- a/install-daemon.sh
+++ b/install-daemon.sh
@@ -22,7 +22,7 @@ fi
 CURLPATH="$(command -v curl)"
 if [ -z "$CURLPATH" ]; then
     echo "* curl is required in order for this script to work."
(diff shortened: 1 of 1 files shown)


VilhelmPrytz:
This is related to #26

install-daemon.sh
```diff
@@ -22,7 +22,7 @@ fi
 CURLPATH="$(command -v curl)"
 if [ -z "$CURLPATH" ]; then
     echo "* curl is required in order for this script to work."
-    echo "* install using apt on Debian/Ubuntu or yum on CentOS"
+    echo "* install using apt on Debian/Ubuntu/Zorin or yum on CentOS"
```
VilhelmPrytz: I think people would understand that `apt` is available on Zorin OS as well (not necessary to put here).

PR #35, opened 2020-02-28
Package not installed when function is ran

### LINKED ISSUE #34: "software-properties-common" installed after it's required
**Describe the bug**
Package `software-properties-common` installed after it's required.

**To Reproduce**
Run script.

**Expected behaviour**
Installed before it's required.

- OS: Ubuntu
 - Installation: panel

MrFlacko:
The package software-properties-common isn't installed before the command it provides (add-apt-repository) is ran.
The fix basically installs the necessary package. Though it is suggestion that this package is installed another way, or since the command 'apt install software-properties-common -y' is already within the script, that command should be ran before this script.

Though for now, this fixes the error message the start of running the script saying "Error: add-apt-repository. command not found"

Fixes #34

### DIFF
--- a/install-panel.sh
+++ b/install-panel.sh
@@ -479,8 +479,8 @@ function centos8_dep {
 #################################
 
 function ubuntu_universedep {
-  #Probably should change this, this is more of a bandaid fix for this 
-  #This function running before software-properties-common is installed
+  # Probably should change this, this is more of a bandaid fix for this 
+  # This function is ran before software-properties-common is installed
   apt update -y
   apt install software-properties-common -y
 
--- a/install-panel.sh
+++ b/install-panel.sh
@@ -479,6 +479,11 @@ function centos8_dep {
 #################################
 
 function ubuntu_universedep {
+  #Probably should change this, this is more of a bandaid fix for this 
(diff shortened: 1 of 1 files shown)


install-panel.sh
```diff
@@ -479,6 +479,11 @@ function centos8_dep {
 #################################
 
 function ubuntu_universedep {
+  #Probably should change this, this is more of a bandaid fix for this 
+  #This function running before software-properties-common is installed
+  apt update -y
```
VilhelmPrytz: Is the `apt update -y` required here?

MrFlacko: Not sure, never tested.
I just put it in there for good measure

MrFlacko: Looks like it is needed
You have made your own function to update and upgrade the debian package manager

though at the start of the function "perform_install" the "ubuntu_universedep" is called before the first time "apt_update" function is called.

So to get it installed my way, it has to have the apt update, as some people install this on old systems that haven't been updated in a long time. So I suggest leaving it in to keep my suggested fix for good measure

VilhelmPrytz: 👍

PR #37, opened 2020-02-29
Added Optional UFW Install and Setup.

MrFlacko:
Added an Opt-In UFW install and setup.
The functions just ask you if you are going to be installed the daemon on the machine, and depending on your answer it will install and setup UFW to allow the required ports, as well as run a little function for you to add more ports.

I also changed some of the wording in the script.

This has been tested with Ubuntu 18.04, though should work on 16.04 and not interfere with other installs

### DIFF
--- a/install-panel.sh
+++ b/install-panel.sh
@@ -577,6 +577,7 @@ function perform_install {
     configure
     insert_cronjob
     install_pteroq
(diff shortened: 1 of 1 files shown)


VilhelmPrytz:
Thank you for your contribution, this is superseded by #38

PR #38, opened 2020-02-29
Let's Encrypt SSL, UFW and suggested daemon install.

VilhelmPrytz:
- Optional: setup Let's Encrypt SSL automatically
- Optional: configure UFW.
- Suggest to the user: install the daemon on the same machine.

Thanks to @MrFlacko for the initial work of this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,12 +81,7 @@ apt install -y ufw
 
 #### Panel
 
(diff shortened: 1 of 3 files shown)


VilhelmPrytz:
Features implemented. Will conduct further testing before merging.

PR #62, opened 2020-07-13
 More compatibility checks

Linux1231233:
I added **architecture checking**, **virtualization type checking** and **checking for unsupported kernel parameters (xxxx)**. (Check out the code for any spelling mistakes).

### DIFF
--- a/install-daemon.sh
+++ b/install-daemon.sh
@@ -115,9 +115,14 @@ function detect_distro {
 function check_os_comp {
   MACHINE_TYPE=$(uname -m)
   if [ "${MACHINE_TYPE}" != "x86_64" ]; then # check the architecture
-      print_error "Detected unsupported architecture $MACHINE_TYPE"
-      print_error "Use 64 bit architecture(x86_64)"
-      exit 1
+      print_warning "Detected architecture $MACHINE_TYPE"
+	   print_warning "Using any other architecture then 64 bit(x86_64) may (and will) cause problems."
+      echo -e -n  "* Are you sure you want to proceed? (y/N):"
+	   read -r choise
+      if [[ ! "$choise" =~ [Yy] ]]; then
+	    print_error "Installation aborted!"
+	    exit 1
+	  fi
   fi
   if [ "$OS" == "ubuntu" ]; then
     if [ "$OS_VER_MAJOR" == "16" ]; then
@@ -162,7 +167,7 @@ function check_os_comp {
     exit 1
   fi
   if [ "$OS" == "debian" ] || [ "$OS" == "ubuntu" ] || [ "$OS" == "zorin" ]; then
-    apt-get update
+    apt-get -y update
 
     # install virt-what
     apt-get install -y virt-what
@@ -185,12 +190,11 @@ function check_os_comp {
     exit 1
   fi
   echo -e "* ${COLOR_RED}Note${COLOR_NC} Detecting virtualization"
+  
   print_brake 70
   
   virt_serv=$(virt-what)
-  if [ "$virt_serv" = "" ]; then
-      print_warning "No virtualization detected"
-  else
+  if [ "$virt_serv" != "" ]; then
       print_warning "Virtualization: $virt_serv detected."
   fi
   
--- a/install-daemon.sh
+++ b/install-daemon.sh
@@ -170,7 +170,7 @@ function check_os_comp {
    elif [ "$OS" == "centos" ]; then
 	
     if [ "$OS_VER_MAJOR" == "7" ]; then
-      yum install -y update
+      yum -y update
 
       # install virt-what
       yum -y install virt-what
--- a/install-daemon.sh
+++ b/install-daemon.sh
@@ -115,15 +115,18 @@ function detect_distro {
 function check_os_comp {
   MACHINE_TYPE=$(uname -m)
   if [ "${MACHINE_TYPE}" != "x86_64" ]; then # check the architecture
-      print_warning "Detected architecture $MACHINE_TYPE"
-	   print_warning "Using any other architecture then 64 bit(x86_64) may (and will) cause problems."
-           echo -e -n  "* Are you sure you want to proceed? (y/N):"
-	   read -r choice
-      if [[ ! "$choice" =~ [Yy] ]]; then
-	    print_error "Installation aborted!"
-	    exit 1
-	  fi
+    print_warning "Detected architecture $MACHINE_TYPE"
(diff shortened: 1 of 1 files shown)


install-daemon.sh
```diff
@@ -113,6 +113,12 @@ function detect_distro {
 }
 
 function check_os_comp {
+  MACHINE_TYPE=$(uname -m)
+  if [ "${MACHINE_TYPE}" != "x86_64" ]; then # check the architecture
```
vilhelmprytz, line 117: Does Pterodactyl (or any of the dependencies) only support 64-bit?

install-daemon.sh
```diff
@@ -113,6 +113,12 @@ function detect_distro {
...
+      print_error "Detected unsupported architecture $MACHINE_TYPE"
+	  print_error "Use 64 bit architecture(x86_64)"
```
vilhelmprytz, line 119: Is this indented differently than the line above or is my phone just showing it weirdly?

install-daemon.sh
```diff
@@ -155,6 +161,50 @@ function check_os_comp {
...
+  if [ "$OS" == "debian" ] || [ "$OS" == "ubuntu" ] || [ "$OS" == "zorin" ]; then
+    apt-get update
+
+    # install virt-what
+    apt-get install -y virt-what
+	
+   elif [ "$OS" == "centos" ]; then
+	
+    if [ "$OS_VER_MAJOR" == "7" ]; then
+      yum install -y update
```
vilhelmprytz, line 173: Hmm, is this supposed to be `yum-y update`?

install-daemon.sh
Linux1231233: This may be my text's editor's fault. In my code it doesn't show up like this.

install-daemon.sh
Linux1231233: No, it may support different architectures, but that wasn't my point. The docker images ,which egg creators use, are mostly based on 64 bit (in my experience). And the commands inside the docker containers also may not be supported by other architectures. But adding a bypass for those who want to take the risk would solve this.

install-daemon.sh
Linux1231233: Yes correct. Another typo.😊

install-daemon.sh
Linux1231233: Fixed in github.

install-daemon.sh
vilhelmprytz: Very good point, closing conversation.

install-daemon.sh
```diff
@@ -155,6 +161,50 @@ function check_os_comp {
...
+  
+  if [ "$virt_serv" == "openvz" ] || [ "$virt_serv" == "lxc" ] ; then # add more virtualization types which are not supported
+      print_warning "Unsupported type of virtualization detected. Please consult with your hosting provider whether your server can run Docker or not. Proceed at your own risk."
+      print_error "Installation aborted!"
+      exit 1
+  fi
+  if uname -r | grep -q "xxxx"; then
+      print_error "Unsupported kernel detected."
+      exit 1
+  fi
```
vilhelmprytz, line 207: This section is performed after the initial welcome message is printed meaning the whole output will just disappear since  `apt`/`dnf`/`yum` do print a lot of information when running. Can `check_os_comp` on line 446 be moved to after line 432?

install-daemon.sh
```diff
@@ -155,6 +161,50 @@ function check_os_comp {
     print_error "Unsupported OS"
     exit 1
   fi
+  if [ "$OS" == "debian" ] || [ "$OS" == "ubuntu" ] || [ "$OS" == "zorin" ]; then
+    apt-get update
+
+    # install virt-what
+    apt-get install -y virt-what
```
vilhelmprytz, line 168: While I do not oppose the use of `virt-what`, are you aware of any other method of detecting virtualization that do not require the extra dependency? (maybe in `/proc/cpuinfo`?)

install-daemon.sh
Linux1231233: Yes, it can be moved without a problem. 👍

install-daemon.sh
Linux1231233: Should I still make a bypass?

install-daemon.sh
Linux1231233: I thought about that, but when trying out the commands like `lscpu`, `sudo dmidecode -s system-manufacturer` I didn't get persistent results on my vm.

vilhelmprytz: Hmm, alright. I'll let it go for now.

vilhelmprytz:
virtwhat

Hmm, this looks a little strange. Is it supposed to do so?

Linux1231233:
I'm not sure how you managed to get that. Very strange tho. It is some kind of issue with print_warning or smth.

vilhelmprytz:
@Linux1231233 Since most people do not use any virtualization when using the daemon (Wings), it would be great if it can handle that case (no messages printed what so ever).

Also, can you mute the output of the `dnf`, `yum` and `apt` commands (and just display some message when they are started?

Linux1231233:
Fixed the print_brake issue, and silenced the output of `dnf`, `yum` and `apt`.

Linux1231233:
Should I redirect outputs to `/dev/null` or just leave warnings and errors?

vilhelmprytz:
@Linux1231233 Perhaps print errors and ignore the rest? What do you think?

Linux1231233:
This is what it is doing right now I think. But with apt-get I used qq > `/dev/null` for dpkg silencing. So I think this means that no errors from apt-get.

install-daemon.sh
```diff
@@ -113,6 +113,17 @@ function detect_distro {
 }
 
 function check_os_comp {
+  MACHINE_TYPE=$(uname -m)
+  if [ "${MACHINE_TYPE}" != "x86_64" ]; then # check the architecture
+      print_warning "Detected architecture $MACHINE_TYPE"
+	   print_warning "Using any other architecture then 64 bit(x86_64) may (and will) cause problems."
+      echo -e -n  "* Are you sure you want to proceed? (y/N):"
+	   read -r choise
```
vilhelmprytz, line 121: I believe it's spelled "choice" 🙂

vilhelmprytz: This line and the two lines above appear differently indented on GitHub. Is it just in GitHub?

Linux1231233: Hmmmmn I need to change my text editor. Also yeah choice another typo.

Issue #61, opened 2020-06-24
Script stops when certbot cannon issue a certificate.


arantesxyz:
**Describe the bug**
When installing the panel, almost at the end of the script, if for some reason certbot fails to create the ssl the script just exit.

**To Reproduce**
Steps to reproduce the behaviour:
1. Run the script
2. Make certbot fails the cert
3. See error.

**Expected behaviour**
If certbot fails, the script should ask if the user still wants to continue the installation with ssl enabled or if he want it to fall back to http.

(I would love the option to continue assuming ssl, because for me certbot will always fail - not a bug, that's my server configuration)

**OS (please complete the following information):**
 - OS: Ubuntu 20.04
 - Installation: panel

vilhelmprytz:
This is for the Pterodactyl 1.0 version of the script, correct? (#52)

arantesxyz:
> This is for the Pterodactyl 1.0 version of the script, correct? (#52)

Correct!

vilhelmprytz:
Will be closed when #52 merges. Thank you @Linux1231233 for the contribution!

PR #68, opened 2020-07-16
Adding support to bypass letsencrypt after failing it

Linux1231233:
Issue #61 fix. Thinking about adding more functionality :)

### DIFF
--- a/install-panel.sh
+++ b/install-panel.sh
@@ -603,12 +603,9 @@ function debian_based_letsencrypt {
 
   systemctl stop nginx
 
-  echo -e "\nMake sure you choose Option 1, and create a Standalone Web Server during the certificate"
   certbot certonly --standalone -d "$FQDN" || FAILED=true
 
-  if [ -d "/etc/letsencrypt/live/$FQDN/" ] || [ "$FAILED" == true ]; then
-    systemctl restart nginx
-  else
+  if [ ! -d "/etc/letsencrypt/live/$FQDN/" ] || [ "$FAILED" == true ]; then
     print_warning "The process of obtaining a Let's Encrypt certificate failed!"
     echo "* Still assume SSL? (y/N): "
     read -r CONFIGURE_SSL
@@ -618,6 +615,8 @@ function debian_based_letsencrypt {
     else
       ASSUME_SSL=false
     fi
+  else 
+    systemctl restart nginx
   fi
 }
 
--- a/install-panel.sh
+++ b/install-panel.sh
@@ -610,7 +610,7 @@ function debian_based_letsencrypt {
     systemctl restart nginx
   else
     print_warning "The process of obtaining a Let's Encrypt certificate failed!"
-    echo "* Still assume ssl? (y/N): "
+    echo "* Still assume SSL? (y/N): "
     read -r CONFIGURE_SSL
 
     if [[ "$CONFIGURE_SSL" =~ [Yy] ]]; then
--- a/install-panel.sh
+++ b/install-panel.sh
@@ -609,7 +609,7 @@ function debian_based_letsencrypt {
   if [ -d "/etc/letsencrypt/live/$FQDN/" ]; then
     systemctl restart nginx
   else
-    print_warning "Obtaining a certificate failed."
+    print_warning "The process of obtaining a Let's Encrypt certificate failed!"
     echo "* Still assume ssl? (y/N): "
     read -r CONFIGURE_SSL
 
--- a/install-panel.sh
+++ b/install-panel.sh
@@ -603,6 +603,7 @@ function debian_based_letsencrypt {
 
   systemctl stop nginx
 
+  FAILED=false
   certbot certonly --standalone -d "$FQDN" || FAILED=true
 
   if [ ! -d "/etc/letsencrypt/live/$FQDN/" ] || [ "$FAILED" == true ]; then
--- a/install-panel.sh
+++ b/install-panel.sh
@@ -606,7 +606,7 @@ function debian_based_letsencrypt {
   echo -e "\nMake sure you choose Option 1, and create a Standalone Web Server during the certificate"
   certbot certonly --standalone -d "$FQDN"
 
-  if [ -d "/etc/letsencrypt/live/$FQDN/" ]; then
+  if [ -d "/etc/letsencrypt/live/$FQDN/" ] || [ "$FAILED" == true ]; then
     systemctl restart nginx
   else
     print_warning "The process of obtaining a Let's Encrypt certificate failed!"
--- a/install-panel.sh
+++ b/install-panel.sh
@@ -602,9 +602,21 @@ function debian_based_letsencrypt {
   systemctl stop nginx
 
   echo -e "\nMake sure you choose Option 1, and create a Standalone Web Server during the certificate"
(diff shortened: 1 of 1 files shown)


install-panel.sh
```diff
@@ -604,9 +604,21 @@ function debian_based_letsencrypt {
   systemctl stop nginx
 
   echo -e "\nMake sure you choose Option 1, and create a Standalone Web Server during the certificate"
-  certbot certonly -d "$FQDN"
+  certbot certonly --standalone -d "$FQDN"
```
vilhelmprytz, line 607: I like the way you check if the certificate exists or not, but this line will cause the script to exit since `set -e` is run on line 3 (which means script exists immediately if any commands return other exit codes than 0).

install-panel.sh
```diff
@@ -604,9 +604,21 @@ function debian_based_letsencrypt {
...
 
-  systemctl restart nginx
+  if [ -d "/etc/letsencrypt/live/$FQDN/" ]; then
+    systemctl restart nginx
+  else
+    print_warning "Obtaining a certificate failed."
```
vilhelmprytz, line 612:
```suggestion
    print_warning "The process of obtaining a Let's Encrypt certificate failed!"
```

Perhaps something like this? Language police here, sorry .... :slightly_smiling_face:

install-panel.sh
```diff
@@ -604,9 +604,21 @@ function debian_based_letsencrypt {
...
+    echo "* Still assume ssl? (y/N): "
```
vilhelmprytz, line 613:
```suggestion
    echo "* Still assume SSL? (y/N): "
```

Linux1231233:
I was thinking about adding a --agree-tos and -n and --register-unsafely-without-email ,but that would kinda give no control over the certificate.

install-panel.sh
Linux1231233: I am not that keen on bash :(
I would think I need something like `return 0`?

vilhelmprytz:
```suggestion
  certbot certonly --standalone -d "$FQDN" || FAILED=true
```

I believe you can do something like this. The `||` is like "else" in an "if-else"-statement. So, in this case, it will set the variable `FAILED` to true (which may then be used on line 609 to check). See the suggestion below.

Again, I am not completely sure this works in this instance. But a simple test like this can verify that it *should* work.

```bash
#!/bin/bash

set -e

failed=false
command_that_does_not_exist_and_therefore_will_fail || failed=true

echo "$failed"
```

If I run this, it prints `true` since the command does not exist and fails. If I then replace command_that_does_not_exist_and_therefore_will_fail` with a command that I know works, then the script prints `false`.

install-panel.sh
```diff
@@ -604,9 +604,21 @@ function debian_based_letsencrypt {
   systemctl stop nginx
 
   echo -e "\nMake sure you choose Option 1, and create a Standalone Web Server during the certificate"
-  certbot certonly -d "$FQDN"
+  certbot certonly --standalone -d "$FQDN"
 
-  systemctl restart nginx
+  if [ -d "/etc/letsencrypt/live/$FQDN/" ]; then
```
vilhelmprytz, line 609:
```suggestion
  if [ -d "/etc/letsencrypt/live/$FQDN/" ] || [ "$FAILED" == true ]; then
```

vilhelmprytz:
> I was thinking about adding a --agree-tos and -n and --register-unsafely-without-email ,but that would kinda give no control over the certificate.

Perhaps it's best to leave those to the user? I reckon it is good practice to provide your email address since it will send notifications when the certificates are about to expire (if auto-renewal fails). :slightly_smiling_face:

Linux1231233:
> Perhaps it's best to leave those to the user? I reckon it is good practice to provide your email address since it will send notifications when the certificates are about to expire (if auto-renewal fails). 🙂
Yeah that is why I didn't add it to the PR. But `--agree-tos` still would be nice. :)

vilhelmprytz:
> > Perhaps it's best to leave those to the user? I reckon it is good practice to provide your email address since it will send notifications when the certificates are about to expire (if auto-renewal fails). 
> 
> Yeah that is why I didn't add it to the PR. But `--agree-tos` still would be nice. :)

Yeah, I agree it's user-friendly and quite convenient but perhaps the user should be aware of what TOS they are accepting?

Linux1231233:
> Yeah, I agree it's user-friendly and quite convenient but perhaps the user should be aware of what TOS they are accepting?

Yes agreed.

vilhelmprytz:
> > Yeah, I agree it's user-friendly and quite convenient but perhaps the user should be aware of what TOS they are accepting?
> 
> Yes agreed.

Great! This should be fairly easy to test, just use an IP and try to get a certificate should trigger the fail. If you are happy with the changes, mark the PR as ready for review and I'll test it and merge it! 👍

Linux1231233:
I tested it already without running the entire script

Going to mark as ready 👍

Linux1231233:
Will there by a problem if there is no FAILED=false defined before the letsencrypt function?

install-panel.sh
```diff
@@ -603,10 +603,21 @@ function debian_based_letsencrypt {
 
   systemctl stop nginx
 
-  echo -e "\nMake sure you choose Option 1, and create a Standalone Web Server during the certificate"
-  certbot certonly -d "$FQDN"
+  certbot certonly --standalone -d "$FQDN" || FAILED=true
```
vilhelmprytz, line 606:
```suggestion
  FAILED=false
  certbot certonly --standalone -d "$FQDN" || FAILED=true
```
In reference to https://github.com/vilhelmprytz/pterodactyl-installer/pull/68#issuecomment-662536826

I think you are very much correct, this might work?

install-panel.sh
```diff
@@ -603,10 +603,22 @@ function debian_based_letsencrypt {
...
-  systemctl restart nginx
+  if [ ! -d "/etc/letsencrypt/live/$FQDN/" ] || [ "$FAILED" == true ]; then
+    print_warning "The process of obtaining a Let's Encrypt certificate failed!"
+    echo "* Still assume SSL? (y/N): "
+    read -r CONFIGURE_SSL
+
+    if [[ "$CONFIGURE_SSL" =~ [Yy] ]]; then
+      ASSUME_SSL=true
+    else
+      ASSUME_SSL=false
```
vilhelmprytz, line 617:
```suggestion
      ASSUME_SSL=false
      CONFIGURE_LETSENCRYPT=false
```

install-panel.sh
```diff
@@ -603,10 +603,22 @@ function debian_based_letsencrypt {
...
+  certbot certonly --standalone -d "$FQDN" || FAILED=true
 
-  systemctl restart nginx
+  if [ ! -d "/etc/letsencrypt/live/$FQDN/" ] || [ "$FAILED" == true ]; then
+    print_warning "The process of obtaining a Let's Encrypt certificate failed!"
+    echo "* Still assume SSL? (y/N): "
+    read -r CONFIGURE_SSL
+
+    if [[ "$CONFIGURE_SSL" =~ [Yy] ]]; then
+      ASSUME_SSL=true
```
vilhelmprytz, line 615:
```suggestion
      ASSUME_SSL=true
      CONFIGURE_LETSENCRYPT=false
```

PR #52, opened 2020-04-19
Pterodactyl 1.0

vilhelmprytz:
Scripts adjusted for Pterodactyl 1.0 which is currently in beta. Will merge when Pterodactyl 1.0 releases!

Before merge: Revert the download and config URLs!!

**Approximate to-do list**

- [x] Try the script for all supported panel installations
    - [x] Ubuntu 18
    - [x] Ubuntu 20
    - [x] Debian 9
    - [x] Debian 10
    - [x] CentOS 7
    - [x] CentOS 8
- [x] Try the script for all supported wings installations
    - [x] Ubuntu 18
    - [x] Ubuntu 20
    - [x] Debian 9
    - [x] Debian 10
    - [x] CentOS 7
    - [x] CentOS 8
- [x] Update Wings to accommodate for the new default install path
- [x] Support CentOS 8
- [x] Initial rewrite for supporting panel
- [x] Initial rewrite for supporting Wings
- [x] wings: support Ubuntu 20.04
- [x] panel: support Ubuntu 20.04

Fixes #41 
Fixes #58 
Fixes #61 
Fixes #63
Fixes #64
Fixes #75
Fixes #86
Fixes #89

PR #70, opened 2020-07-17
Show progress every once in a while

Sam1370:
This pull request shows the progress of the installation every once in a while. This isn't perfect by any means, although I think it could be mildly useful. 

Ideally, this would be a progress bar that is persistent and stays at the bottom of the screen, and updates when a function is called, although I do not know how to implement that.

### DIFF
--- a/install-panel.sh
+++ b/install-panel.sh
@@ -14,6 +14,18 @@
 #                                                                           #
 #############################################################################
 
+# Load progressbar
+source <(curl -s https://raw.githubusercontent.com/roddhjav/progressbar/master/progressbar.sh) || exit 1
+
+# Initialize percent done
+percentdone=0
+
+# Progress bar function
+display_progress() {                                                                                                                                         
+        progressbar "PANEL INSTALLATION COMPLETION" $percentdone 100                                                                                                    
+        echo                                                                                                                                                 
+}
+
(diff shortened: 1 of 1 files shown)


Sam1370:
Here is a demonstration:
https://www.youtube.com/watch?v=OZSUEVTSmpM

vilhelmprytz:
Thank you for your contribution! I really like the idea of displaying some form of progress bar once a while. Is it possible to create a progress bar without using a third-party script? Or perhaps make a copy of the source script into this project? (license and copyright notices from the original author kept, of course!)

Also, because of the many changes in #52 for the pterodactyl 1.0 version of the script, are you able to write these changes using `pterodactyl-1.0` as your base branch (e.g. start from 1.0 modified version of the script)? In that way, your work doesn't have to be rewritten for 1.0 :slightly_smiling_face:

vilhelmprytz:
Perhaps we can find a middle-point here, copying something from StackOverflow isn't exactly great when it comes to copyright and such, is it possible to use `tput` and write something similar?

Linux1231233:
At least I can write anything like this. It is too hard for me :)

vilhelmprytz:
Since this has failing builds, merge conflicts, and there is another PR solving the same issue (#87), I am closing this.

PR #71, opened 2020-07-22
Fixing nginx stopping the script

Linux1231233:
When you choose ASSUME_SSL true, and LETSENCRYPT false the nginx will fail at the end of the script. This fixes that.

### DIFF
--- a/install-panel.sh
+++ b/install-panel.sh
@@ -666,7 +666,12 @@ function configure_nginx {
   fi
 
   # restart nginx
-  systemctl restart nginx
+  if [ "$CONFIGURE_LETSENCRYPT" == true ] || { [ "$CONFIGURE_LETSENCRYPT" == false ] && [ "$ASSUME_SSL" == false ]; }; then
+    systemctl restart nginx
+  else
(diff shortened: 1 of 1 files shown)


vilhelmprytz:
Thank you for the contribution! I'll look into it and review it ASAP!

Linux1231233:
I saw you prefer one liners  so there it is:
`{ [ "$CONFIGURE_LETSENCRYPT" == true ] || { [ "$CONFIGURE_LETSENCRYPT" == false ] && [ "$ASSUME_SSL" == false ]; }; } && systemctl restart nginx`

vilhelmprytz:
> I saw you prefer one liners  so there it is:
> `{ [ "$CONFIGURE_LETSENCRYPT" == true ] || { [ "$CONFIGURE_LETSENCRYPT" == false ] && [ "$ASSUME_SSL" == false ]; }; } && systemctl restart nginx`

Wow, that's very much shorter. I reckon it's actually more readable with the one you initially wrote. What do you think?

Linux1231233:
That is exactly why I didn't push it to the PR and just commented.

PR #72, opened 2020-07-24
Added master script 

### LINKED ISSUE #66: Add “master” script (entrypoint for the other scripts)
Currently, you have to run different commands for each script: panel, daemon, both, the 1.0 versions, etc. With the addition of a “master” script, one script can be ran that will allow you to choose what thing you want installed.

One thing that may have to be removed to eliminate confusion: the option to install the daemon at the end of the panel script.

Linux1231233:
This is still very work in progress. Merge this after merging #52 (or merging parts of it) to master.

@vilhelmprytz Write out all the things which should be done in this master script and not in the install scripts (for example checking for sudo). I know the code is very bad :(

Fixes #66


vilhelmprytz:
I did some changes, what do you think?

Linux1231233:
Some changes? This a whole rework. I think it is great. :)

Linux1231233:
Also, don't forget to change the README.md

PR #73, opened 2020-07-25
Adding apache support

Linux1231233:
Added apache support. Untested as of right now.

### DIFF
--- a/configs/apache.conf
+++ b/configs/apache.conf
@@ -1,0 +1,10 @@
+<VirtualHost *:80>
+  ServerName <domain>
+  DocumentRoot "/var/www/pterodactyl/public"
+  AllowEncodedSlashes On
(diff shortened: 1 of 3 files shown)


vilhelmprytz:
Thank you for the contribution! While I think supporting Apache is good when it comes to general compatibility, I think most people benefit from sticking to Nginx (generally works and performs better IMO). And supporting Apache also means a ton of work on the script, especially when it comes to maintaining it in the future. Are you sure you want to continue testing this? I am not sure I will have time to maintain what is essentially a doubled amount of machine configurations. When a new release of Pterodactyl comes out or a new version of a supported distributions comes out, that means quite a lot of work.

Linux1231233:
Ok, no problem going to close the PR 👍

PR #77, opened 2020-07-29
panel: remove TLSv1.3 ssl_protocol in nginx config on debian 8/9

### LINKED ISSUE #76: Debian 8/9 nginx version too old
**Describe the bug**

When installing nginx from default repos, it gets quite an old version of it (1.10.3). It has a lot of drawbacks (for e. g. TLSv1.3 isn't supported).

**To Reproduce**

1. Run the panel installation script.
2. Run `nginx -t`
3. See syntax error TLSv1.3  

**Expected behaviour**

It should install a newer version of nginx which supports new features.

 - OS: Debian 8/9
 - Installation: panel

vilhelmprytz:
Fixes #76

### DIFF
--- a/install-panel.sh
+++ b/install-panel.sh
@@ -583,6 +583,14 @@ function configure_nginx {
       # replace all <php_socket> places with correct socket "path"
       sed -i -e "s@<php_socket>@${PHP_SOCKET}@g" /etc/nginx/sites-available/pterodactyl.conf
 
+      # on debian 8/9, TLS v1.3 is not supported (see #76)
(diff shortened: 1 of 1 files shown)


vilhelmprytz:
@Linux1231233 Are you able to test if this is better?

Linux1231233:
Not right now but I will do it asap.

vilhelmprytz:
Alright. I tested it locally and it does indeed remove the specified part correctly. I doubt the other parts of this PR (e.g. checking whether the user is on Debian 8/9 or not) is incorrect so I'll just merge it :+1: Thank you for reporting this bug!

PR #84, opened 2020-08-05
Add template and cleanup code

### LINKED ISSUE #75: Adding the .env template and automatically filling it in.
Lots of people don't know what to select in the `php artisan p:environment:setup`, `php artisan p:environment:database`, `php artisan p:environment:mail` commands.

Creating a .env file template, and replacing the crucial variables in it.

Not applicable.

Linux1231233:
Fixes #75
Adding a .env template and filling it in using sed.

Also purging any apache in the code, removing support for zorin. (cleaning up the code)


vilhelmprytz:
Thanks for the contribution, I'll look at it more closely ASAP! :+1:

Linux1231233:
If you really want to have no prompts after the initial one's we would need to add Something like 
"Do you agree to letsencrypt TOS?"(link to TOS). 
If no we fallback to http. Then we add `--agree-tos` to certbot.
But there will be left one more prompt about EFF. I think we can savely just answer No to it because it is not related to letsencrypt's creation of certificates. Then `--no-eff-email` and there is no prompts at all in certbot part.

Linux1231233:
Also if you want no prompts we would need to do the `php artisan p:environment:mail` and `php artisan p:user:make` at the end of the script. It wouldn't be possible to do that a the start. (Mail would be possible to autofill, but really hard for the user. I have added a bypass if you don't want to setup the mail sending). I hope you get what I mean,

vilhelmprytz:
> Also if you want no prompts we would need to do the `php artisan p:environment:mail` and `php artisan p:user:make` at the end of the script. It wouldn't be possible to do that a the start. (Mail would be possible to autofill, but really hard for the user. I have added a bypass if you don't want to setup the mail sending). I hope you get what I mean,
> 

@Linux1231233 Sorry for the delay. Yes, I understand what you mean. The primary reason to put all configuration prompts in the beginning of the script is to avoid the script stalling and waiting for user input. Since the `mail` and `user` prompts occur towards the very end of the script, it isn't too bad since the installation process is pretty much completed after those two steps. Reviewing this is on my to-do list, we'll get back to you on that.

Linux1231233:
> If you really want to have no prompts after the initial one's we would need to add Something like
"Do you agree to letsencrypt TOS?"(link to TOS).
If no we fallback to http. Then we add --agree-tos to certbot.
But there will be left one more prompt about EFF. I think we can savely just answer No to it because it is not related to letsencrypt's creation of certificates. Then --no-eff-email and there is no prompts at all in certbot part.

What about this? Or is this not needed to change, bcs it is at the end?

Sam1370:
> If you really want to have no prompts after the initial one's we would need to add Something like
> "Do you agree to letsencrypt TOS?"(link to TOS).
> If no we fallback to http. Then we add `--agree-tos` to certbot.

Think it should be made clearer that denying TOS will result in no HTTPS.

vilhelmprytz:
> > If you really want to have no prompts after the initial one's we would need to add Something like
> > "Do you agree to letsencrypt TOS?"(link to TOS).
> > If no we fallback to http. Then we add `--agree-tos` to certbot.
> 
> I think it should be made clearer that denying TOS will result in no HTTPS.

@Sam1370 But on the other hand, the script now handles the `certbot` command failing.

vilhelmprytz:
> > If you really want to have no prompts after the initial one's we would need to add Something like
> > "Do you agree to letsencrypt TOS?"(link to TOS).
> > If no we fallback to http. Then we add --agree-tos to certbot.
> > But there will be left one more prompt about EFF. I think we can savely just answer No to it because it is not related to letsencrypt's creation of certificates. Then --no-eff-email and there is no prompts at all in certbot part.
> 
> What about this? Or is this not needed to change, bcs it is at the end?

@Linux1231233 Hmm, I don't know really. I think the `--no-eff-email` is not a good idea since it sends emails when the certificate is about to expire (like if auto-renewal failed). And if that is to be left in, the `--agree-tos` is really also redundant.

Linux1231233:
@vilhelmprytz `--no-eff-email` is for the eff forum or whatever it is. It still sends emails about renewals.

vilhelmprytz:
> @vilhelmprytz `--no-eff-email` is for the eff forum or whatever it is. It still sends emails about renewals.

@Linux1231233 So it will still prompt the user to enter an email address for certificate renewal notifications even when `--no-eff-email` is specified?

Linux1231233:
@vilhelmprytz We have the email address already because we ask it when we fill in the .env . So we use `--email $email`. And yes it will send it even if you decline getting messages from that eff forum.

install-panel.sh
```diff
@@ -604,7 +596,7 @@ function debian_based_letsencrypt {
   systemctl stop nginx
 
   FAILED=false
-  certbot certonly --standalone -d "$FQDN" || FAILED=true
+  certbot certonly --email "$email" --standalone -d "$FQDN" || FAILED=true # -q could be added because it already checks if it failed
```
vilhelmprytz, line 599: Good use of `--email`, I agree then you can use `--no-eff-email`. Do it! :+1: Still think the user should accept the TOS since it is a legal document.

install-panel.sh
```diff
@@ -680,10 +672,6 @@ function configure_nginx {
   echo "* nginx configured!"
 }
 
-function configure_apache {
```
vilhelmprytz, line 683: Ah, I had even forgotten that this was here! Good to get it removed! :+1:

install-panel.sh
```diff
@@ -874,6 +845,22 @@ function main {
     [ -z "$MYSQL_PASSWORD" ] && print_error "MySQL password cannot be empty"
   done
 
+  print_brake 88
+  echo "* Please follow the steps below. The installer will ask you for configuration details."
+  print_brake 88
+  echo ""
+
+  echo "* Choose your timezone. e.g. [Europe/Amsterdam]: " # This is really hard without autocomplete
```
vilhelmprytz, line 853: Yes, no autocomplete is a problem really. Is there anything that can be done? I believe Pterodactyl uses the [PHP standard format for timezones](https://www.php.net/manual/en/timezones.php). Perhaps we can manually verify that the timezone is valid somehow?

install-panel.sh
```diff
@@ -874,6 +845,22 @@ function main {
...
+  read -r timezone_input
+
+  [ -z "$timezone_input" ] && timezone="Europe/Amsterdam" || timezone=$timezone_input
+
+  echo "* Provide the email address that will be used to configure LetsEncrypt and Egg author: " # Make it more readable
```
vilhelmprytz, line 858:
```suggestion
  echo "* Provide the email address that will be used to configure Let's Encrypt and Pterodactyl: "
```

I am aware the `Egg author` is a more correct and specific term, but perhaps it's okay to just say it's an email that Pterodactyl uses?

install-panel.sh
```diff
@@ -874,6 +845,22 @@ function main {
...
+
+  echo "* Choose your timezone. e.g. [Europe/Amsterdam]: " # This is really hard without autocomplete
+  read -r timezone_input
+
+  [ -z "$timezone_input" ] && timezone="Europe/Amsterdam" || timezone=$timezone_input
+
+  echo "* Provide the email address that will be used to configure LetsEncrypt and Egg author: " # Make it more readable 
+  read -r email
+
+  echo "* Would you like to setup email client for sending emails? (y/N): "
```
vilhelmprytz, line 861:
```suggestion
  echo "* Would you like to set up email credentials so that Pterodactyl can send emails to users? (y/N): "
```

Linux1231233:
So @vilhelmprytz I have found another pickle.

I don't know how you add code like that, but in the mysql setup part in centos section ```mysql_secure_installation```  leads to no function. Am I wrong?

vilhelmprytz:
> So @vilhelmprytz I have found another pickle.
> 
> I don't know how you add code like that, but in the mysql setup part in centos section `mysql_secure_installation` leads to no function. Am I wrong?

Yes and no. Indeed it does not lead to any function within the script but rather an executable from MariaDB/MySQL (see [here](https://mariadb.com/kb/en/mysql_secure_installation/)).

Linux1231233:
Ahh yeah thought that the echoes above meant something. Thanks for clarifying.

PR #85, opened 2020-08-09
Add centos certbot

### LINKED ISSUE #86: Let's Encrypt for CentOS
Centos doesn't have certbot, thus it is not possible to create certificates. 

Add letsencrypt support to centOS

You can create certificates yourself.

**Additional context**
#85 Fixes

Linux1231233:
Adding certbot support for CentOS

Fixes #86


vilhelmprytz:
@Linux1231233 Are you able to open a feature request issue and reference that this PR fixes that issue?

Linux1231233:
Oh yeah forgot that thanks for reminding :)

install-panel.sh
```diff
@@ -590,16 +590,45 @@ function firewall_firewalld {
   echo "* Firewall-cmd installed"
   print_brake 70
 }
+function centos_based_letsencrypt {
+    if [ "$OS_VER_MAJOR" == "7" ]; then
+      # Enable EPEL falowing official guide https://fedoraproject.org/wiki/EPEL
+      yum install https://dl.fedoraproject.org/pub/epel/epel-release-latest-7.noarch.rpm
```
vilhelmprytz, line 596: I am unsure if a conflict occurs here since the function `centos7_dep` also installs some form of epel-release package.

install-panel.sh
```diff
@@ -590,16 +590,45 @@ function firewall_firewalld {
...
+      subscription-manager repos --enable "rhel-*-optional-rpms" --enable "rhel-*-extras-rpms"  --enable "rhel-ha-for-rhel-*-server-rpms"
 
+      # Install certbot 
+      yum install certbot
+    elif [ "$OS_VER_MAJOR" == "8" ]; then
+      # Enable EPEL falowing official guide https://fedoraproject.org/wiki/EPEL
+      dnf install https://dl.fedoraproject.org/pub/epel/epel-release-latest-8.noarch.rpm
```
vilhelmprytz, line 603: I am unsure if a conflict occurs here since the function `centos8_dep` also installs some form of epel-release package.

Linux1231233:
But are they the same EPEL? I would need to do a bit more research.

install-panel.sh
```diff
@@ -590,16 +590,43 @@ function firewall_firewalld {
...
+function centos_based_letsencrypt {
+    if [ "$OS_VER_MAJOR" == "7" ]; then
+      # Enable EPEL folowing official guide https://fedoraproject.org/wiki/EPEL
+      subscription-manager repos --enable "rhel-*-optional-rpms" --enable "rhel-*-extras-rpms"  --enable "rhel-ha-for-rhel-*-server-rpms"
+
+      # Install certbot 
+      yum install certbot
+    elif [ "$OS_VER_MAJOR" == "8" ]; then
+      # Enable EPEL folowing official guide https://fedoraproject.org/wiki/EPEL
+      ARCH=$(/bin/arch)
```
vilhelmprytz, line 602: Why is `ARCH` defined?

install-panel.sh
```diff
@@ -590,16 +590,43 @@ function firewall_firewalld {
...
+      subscription-manager repos --enable "codeready-builder-for-rhel-8-${ARCH}-rpms"
```
vilhelmprytz, line 603: Using `epel-release` seems fairly straight-forward, isn't this overly complicated? [See this guide.](https://www.digitalocean.com/community/tutorials/how-to-secure-apache-with-let-s-encrypt-on-centos-8#step-1-%E2%80%94-installing-the-certbot-let%E2%80%99s-encrypt-client).

Linux1231233: @vilhelmprytz I have removed the installation of EPEL from here. The subscription-manager thing is documented in [https://fedoraproject.org/wiki/EPEL](url) and I haven't found it in the script.

install-panel.sh
Linux1231233: I asked about that in discord you never answered :( 
I would say that we can just hardcode `x86_64` and call it a day. Of Course we can also define it in the variables section :)

vilhelmprytz: > I asked about that in discord you never answered :( 
> I would say that we can just hardcode `x86_64` and call it a day. Of Course we can also define it in the variables section :)

I saw your question on the Discord, my question here is why defining the variable `ARCH` Iis required in the first place?

Linux1231233: The `subscription-manager repos --enable "codeready-builder-for-rhel-8-${ARCH}-rpms"` needs architecture. I would just use x86_64.

vilhelmprytz: > The `subscription-manager repos --enable "codeready-builder-for-rhel-8-${ARCH}-rpms"` needs architecture. I would just use x86_64.

Yes, I understand. I find it overly complicated to activate `epel-release` in this way. Does this not work on CentOS 8?

```bash
sudo dnf install epel-release
sudo dnf install certbot
```

Is there any particular reason for not using the "built-in" epel-release?

Linux1231233: Ok let me start from very very start. So at first I had EPEP release downloaded by certbot start. Then I removed it bcs it was already installed before. Then here it says that you need to enable any recommended repositories. 
image

So in [https://fedoraproject.org/wiki/EPEL#Quickstart](https://fedoraproject.org/wiki/EPEL#Quickstart) it says that some packages may require this 
image

So following the certbot instructions to enable any recommendations I added that to the script. I don't know if the digitalocean page is correct I just read the instructions on the official certbots page. 

[https://certbot.eff.org/lets-encrypt/centosrhel7-other](https://certbot.eff.org/lets-encrypt/centosrhel7-other)
[https://certbot.eff.org/lets-encrypt/centosrhel8-other](https://certbot.eff.org/lets-encrypt/centosrhel8-other)

PS: it may very well work without them. I just followed the instructions.

vilhelmprytz: > Ok let me start from very very start. So at first I had EPEP release downloaded by certbot start. Then I removed it bcs it was already installed before. Then here it says that you need to enable any recommended repositories.
> image
> 
> So in https://fedoraproject.org/wiki/EPEL#Quickstart it says that some packages may require this
> image
> 
> So following the certbot instructions to enable any recommendations I added that to the script. I don't know if the digitalocean page is correct I just read the instructions on the official certbots page.
> 
> https://certbot.eff.org/lets-encrypt/centosrhel7-other
> https://certbot.eff.org/lets-encrypt/centosrhel8-other
> 
> PS: it may very well work without them. I just followed the instructions.

I understand. As far as I know, most of it is only for certain packages that may require it. Can you try without it and if it does not work, then add them?

CentOS 7:
```bash
sudo yum install certbot python2-certbot-nginx
```

CentOS 8:
```bash
sudo dnf install certbot python3-certbot-nginx
```

I am aware of the fact that the certbot documentation recommends enabling the recommended repositories, but I would like to avoid a conflict with the already installed `epel-release` (and I think certbot can do without the extra stuff).

install-panel.sh
```diff
@@ -591,15 +591,28 @@ function firewall_firewalld {
...
-  apt install certbot -y
+function letsencrypt {
+  FAILED=false
+
+  # Install certbot
+  if [ "$OS" == "debian" ] || [ "$OS" == "ubuntu" ]; then
+    apt-get install certbot -y
+  elif [ "$OS" == "centos" ]; then
+    [ "$OS_VER_MAJOR" == "7" ] && yum install certbot
+    [ "$OS_VER_MAJOR" == "8" ] && dnf install certbot
```
vilhelmprytz, line 602: Are you sure the `python2-certbot-nginx` and the `python3-certbot-nginx` are not required (7 and 8 respectively)?

Linux1231233: I think they are only used when using --webroot I think. When using --standalone it's not need because it creates a server itself. Either way we are stopping nginx :)

Linux1231233: Just tested without the `python3-certbot-nginx ` and `python2-certbot-nginx` and it works just fine. Then I guess it was correct I assumed it is only when you don't want to stop nginx.

vilhelmprytz: No, it's for the `certbot --nginx` which may come in handy if the user decides to obtain other certificates (since it works natively with Nginx in the same way certbot "natively" obtains certificate and performs verification with Apache). But this would require other changes in the script so I'll leave it for now, thanks for testing! :+1:

PR #87, opened 2020-08-14
Add progress bar

Linux123123:
Adding a progress bar. Still not finished. Using code snippets from [bash_progress_bar](https://github.com/pollev/bash_progress_bar). Still needs work in the places where we increase percentage. (I stopped adding them after the initial questions)

`block_progress_bar()` is used when waiting for user input
`function draw_progress_bar()` is used when the process is running and percentage needs to be updated

#70

### DIFF
--- a/install-panel.sh
+++ b/install-panel.sh
@@ -41,12 +41,6 @@ if ! [ -x "$(command -v curl)" ]; then
   exit 1
 fi
 
-if ! [ -x "$(command -v tput)" ]; then
-  echo "* curl is required in order for this script to work."
-  echo "* install using apt (Debian and derivatives) or yum/dnf (CentOS)"
-  exit 1
-fi
-
 # define version using information from GitHub
 get_latest_release() {
   curl --silent "https://api.github.com/repos/$1/releases/latest" | # Get latest release from GitHub api
--- a/install-panel.sh
+++ b/install-panel.sh
@@ -97,6 +97,7 @@ CONFIGURE_FIREWALL=false
 
 # shellcheck source=progress_bar.sh
 source <(curl -s https://raw.githubusercontent.com/Linux123123/pterodactyl-installer/pterodactyl-1.0/progress_bar.sh) || exit 1  # revert before merge
+# source progress_bar.sh # for debugging
 
 function print_error {
   COLOR_RED='\033[0;31m'
@@ -665,7 +666,7 @@ function firewall_firewalld {
 
   echo "* Firewall-cmd installed"
   print_brake 70
-  block_progress_bar 31
+  draw_progress_bar 31
 }
 
 function letsencrypt {
@@ -885,7 +886,7 @@ function main {
 
   # Enable progress bar
 
-  enable_bar
+  bar::start
 
   print_brake 70
   echo "* Pterodactyl panel installation script"
@@ -899,11 +900,10 @@ function main {
   print_brake 70
 
   # checks if the system is compatible with this installation script
-  check_os_comp
 
   # set database credentials
   print_brake 72
-  block_progress_bar 0
+  draw_progress_bar 1
 
   echo "* Database configuration."
   echo ""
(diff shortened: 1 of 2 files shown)


install-panel.sh
```diff
@@ -41,6 +41,12 @@ if ! [ -x "$(command -v curl)" ]; then
   exit 1
 fi
 
+if ! [ -x "$(command -v tput)" ]; then
```
vilhelmprytz, line 44: Does `tput` not come as "standard" for any of the supported distros?

install-panel.sh
```diff
@@ -41,6 +41,12 @@ if ! [ -x "$(command -v curl)" ]; then
...
+  echo "* curl is required in order for this script to work."
+  echo "* install using apt (Debian and derivatives) or yum/dnf (CentOS)"
```
vilhelmprytz, line 46: Copied from the `curl` check? :smile:

install-panel.sh
Linux123123: Yes it does I think. Will remove

install-panel.sh
Linux123123: Exactly

vilhelmprytz:
Am I doing something wrong? The script looks like [this](https://i.vilhelmprytz.se/2020/08/2020-08-14_20-22-25.mp4) for me.

Linux123123:
Strange. Maybe you have custom terminal or something. On an ssh server to ubuntu linux it works fine, wsl2 on ubuntu works fine too. This might also be a UTF issue. Try exporting it. Also it seems that towards the end it starts working. 

``` bash
export LC_ALL=en_US.UTF-8
export LANG=en_US.UTF-8
export LANGUAGE=en_US.UTF-8
```

vilhelmprytz:
> Strange. Maybe you have custom terminal or something. On an ssh server to ubuntu linux it works fine, wsl2 on ubuntu works fine too. This might also be an UTF issue. Try exporting it. Also it seems that towards the end it starts working.
> 
> ```shell
> export LC_ALL=en_US.UTF-8
> export LANG=en_US.UTF-8
> export LANGUAGE=en_US.UTF-8
> ```

I do usually use `zsh` over "standard bash" but in the video I linked I am not using zsh, no custom shell. Only a theme on the terminal emulator. I just tried on a Debian 10 machine I have with the same issue, unfortunately.

Linux123123:
I just tried on a machine which has never run this script and got it working. Seems like a country to country issue?

vilhelmprytz:
> I just tried on a machine which has never run this script and got it working. Seems like a country to country issue?

Perhaps, my `$LANG` is `en_US.UTF-8` (without running the `export` commands you sent). I tried with a different terminal emulator with the same result.

Linux123123:
So I have fixed up (I hope) some assignments. This should change the string it gets from `tput lines` to integer. Try again and see if there are any problems :)

vilhelmprytz:
> So I have fixed up (I hope) some assignments. This should change the string it gets from `tput lines` to integer. Try again and see if there are any problems :)

@Linux123123 Thanks, will try it!

vilhelmprytz:
@Linux123123 So, I just tried it on my computer and I still have the same strange visual glitch that I sent in a video for earlier.

Also, would it be possible to move all the functions related to the progress bar into a separate `.sh` file that is then "imported" at the top? By doing this, it will be easier to separate the functions since the ones you copied are licensed under a different license.

Linux123123:
The issue is kinda fixed. When you resize the terminal, it will disappear until the next `drawprogress` function is called. This is the only way this works because the progress bar isn't constantly updating, but rather just redrawing the progress bar. So basically you resize, wait a few seconds for next function call and it is right back there.

PS Talking about `drawprogress` I really need to get the whole script done.

vilhelmprytz:
> The issue is kinda fixed. When you resize the terminal, it will disappear until the next `drawprogress` function is called. This is the only way this works because the progress bar isn't constantly updating, but rather just redrawing the progress bar. So basically you resize, wait a few seconds for next function call and it is right back there.
> 
> PS Talking about `drawprogress` I really need to get the whole script done.

@Linux123123 I just tried it and even though it does sort itself out eventually, there can be several prompts where the questions are barely readable. It's quite a niche scenario, you don't often resize the terminal a lot, but IMO it does look a bit strange. Although, when the progress bar works properly, it does look very nice :+1: :slightly_smiling_face: 

When it comes to the real-world time it takes for the script to finish, most of the time is spent when `apt`/`dnf` is running (and `composer` a bit). `apt`/`dnf` already kinda have their own sort of progress bars which makes me wonder if a separate progress bar is really necessary? I am very much open to discussion on this point.

Linux123123:
Added mosh shell detection! It should now automatically disable progress bar if it finds mosh-server process running.

Linux123123:
I have redone all the code. It now works on mosh (in my testing) and looks nicer.

Linux123123:
Too many issues, closing.

PR #91, opened 2020-08-23
Automatic Let's Encrypt for daemon (wings)

### LINKED ISSUE #58: Automatic Let's Encrypt configuration for daemon
The installation script of the daemon does not create an SSL certificate.
There is still a query missing, which domain should be used incl. installing the certbot and issuing a certificate.
If you install everything on one system, it works without problems. If you separate panel and daemon, it will not work.

If i install certbot and create my own ssl-certificate all works fine :)

vilhelmprytz:
Fixes #58

### DIFF
--- a/install-wings.sh
+++ b/install-wings.sh
@@ -435,6 +435,9 @@ function firewall_ufw {
   ufw allow 8080 > /dev/null
   ufw allow 2022 > /dev/null
 
+  [ "$CONFIGURE_LETSENCRYPT" == true ] && ufw allow http > /dev/null
+  [ "$CONFIGURE_LETSENCRYPT" == true ] && ufw allow https > /dev/null
+
   ufw enable
   ufw status numbered | sed '/v6/d'
 }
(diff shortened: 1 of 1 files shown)


install-wings.sh
```diff
@@ -527,6 +587,27 @@ function main {
...
+  fi
+
+  if [ "$CONFIGURE_LETSENCRYPT" == true ]; then
+    # set FQDN
+    while [ -z "$FQDN" ] || [ -d "/etc/letsencrypt/live/$FQDN/" ]; do
+        echo -n "* Set the FQDN to use for Let's Encrypt (panel.example.com): "
+        read -r FQDN
+
+        [ -z "$FQDN" ] && print_error "FQDN cannot be empty"
+        [ -d "/etc/letsencrypt/live/$FQDN/" ] && print_error "A certificate with this FQDN already exists!"
```
Linux123123, line 601: I think this is not the best place to check that. This should be done in `ask_letsencrypt`. Then of course we would need to check if there is something in `/etc/letsencrypt/live/`, not strictly the FQDN that is intended. Then the script would ask: We have found this `$FOUND_FQDN` do you still want to create certificates?

vilhelmprytz: I initially tried putting them both in `ask_letsencrypt` but somehow the `if` on line 506 (see [this](https://github.com/vilhelmprytz/pterodactyl-installer/pull/91/files#diff-fd99fe6be85e7380f1f97accfc328d59R507)) in combination with the `while` loop would just make it exit. Is there any real problem having the questions here?

Linux123123: Well you get that messages when you are committed to getting certificates. And you would need to exit and restart the script. It would be better to give that information before you press yes to use letsencrypt. Anyway I will also try to get this working.

vilhelmprytz: > Well you get that messages when you are committed to getting certificates. And you would need to exit and restart the script. It would be better to give that information before you press yes to use letsencrypt. Anyway I will also try to get this working.

I agree, but I couldn't get it to work properly without strange shortcuts. Since this is my PR you can leave it to me, thanks for the review. I'll make sure I fix it.

Linux123123:
Ok then, but I have some work already done and I can't really commit it to your branch. So here it is:
```bash
  ufw allow 80 > /dev/null
  ufw allow 443 > /dev/null
```
```bash
  [ "$OS_VER_MAJOR" == "7" ] && yum -y -q update
  [ "$OS_VER_MAJOR" == "7" ] && yum -y -q install firewalld > /dev/null
  [ "$OS_VER_MAJOR" == "8" ] && dnf -y -q update
  [ "$OS_VER_MAJOR" == "8" ] && dnf -y -q install firewalld > /dev/null

  systemctl --now enable firewalld > /dev/null # Enable and start 
  firewall-cmd --add-port 8080/tcp --permanent -q # Port 8080
  firewall-cmd --add-port 2022/tcp --permanent -q # Port 2022
  firewall-cmd --add-port 80/tcp --permanent -q # Port 80
  firewall-cmd --add-port 443/tcp --permanent -q # Port 443
  firewall-cmd --permanent --zone=trusted --change-interface=pterodactyl0 -q
  firewall-cmd --zone=trusted --add-masquerade --permanent
  firewall-cmd --ad-service=ssh --permanent -q # Port 22
  firewall-cmd --reload -q # Enable firewall
```
Both are for the firewall part of the wings script.
1. Add ports 80 443 to `ufw`
2. Cleanup `firewalld` part and also add ports 80 and 443

vilhelmprytz:
> Ok then, but I have some work already done and I can't really commit it to your branch. So here it is:
> 
> ```shell
>   ufw allow 80 > /dev/null
>   ufw allow 443 > /dev/null
> ```
> 
> ```shell
>   [ "$OS_VER_MAJOR" == "7" ] && yum -y -q update
>   [ "$OS_VER_MAJOR" == "7" ] && yum -y -q install firewalld > /dev/null
>   [ "$OS_VER_MAJOR" == "8" ] && dnf -y -q update
>   [ "$OS_VER_MAJOR" == "8" ] && dnf -y -q install firewalld > /dev/null
> 
>   systemctl --now enable firewalld > /dev/null # Enable and start 
>   firewall-cmd --add-port 8080/tcp --permanent -q # Port 8080
>   firewall-cmd --add-port 2022/tcp --permanent -q # Port 2022
>   firewall-cmd --add-port 80/tcp --permanent -q # Port 80
>   firewall-cmd --add-port 443/tcp --permanent -q # Port 443
>   firewall-cmd --permanent --zone=trusted --change-interface=pterodactyl0 -q
>   firewall-cmd --zone=trusted --add-masquerade --permanent
>   firewall-cmd --ad-service=ssh --permanent -q # Port 22
>   firewall-cmd --reload -q # Enable firewall
> ```
> 
> Both are for the firewall part of the wings script.
> 
> 1. Add ports 80 443 to `ufw`
> 2. Cleanup `firewalld` part and also add ports 80 and 443

@Linux123123 Thanks a lot, about the ports, very good point. Cleanup `firewalld` perhaps doesn't belong here but I will include it anyway :slightly_smiling_face:

install-wings.sh
vilhelmprytz: The problem is that if you put the FQDN/email prompts before asking if enabling Let's Encrypt, then the user has to enter potentially redundant information. A solution would be to first ask whether to enable certificates or not then ask for information and if certificate was found, then re-confirm, which also might look bloated. I will see what I can do.

vilhelmprytz: Let me know if I am missing something.

install-wings.sh
```diff
@@ -458,11 +482,27 @@ function perform_install {
...
   # return true if script has made it this far
   return 0
 }
 
+ask_letsencrypt() {
+  if [ "$CONFIGURE_UFW" == false ] && [ "$CONFIGURE_FIREWALL_CMD" == false ]; then
+    print_warning "Let's Encrypt requires port 80/443 to be opened! You have opted out of the automatic firewall configuration; use this at your own risk (if port 80/443 is closed, the script will fail)!"
+  fi
+
+  print_warning "You cannot use Let's Encrypt with your hostname as an IP address! It must be a FQDN (e.g. panel.example.org)."
```
Linux123123, line 496:
```suggestion
  print_warning "You cannot use Let's Encrypt with your hostname as an IP address! It must be a FQDN (e.g. node.example.org)."
```
We should change this because people may be confused that it is panel and that they are doing a bad thing.

Also I was forced to write this review.

install-wings.sh
```diff
@@ -527,6 +581,38 @@ function main {
     if [[ "$CONFIRM_FIREWALL_CMD" =~ [Yy] ]]; then
       CONFIGURE_FIREWALL_CMD=true
     fi
+
+    ask_letsencrypt
+  fi
+
+  if [ "$CONFIGURE_LETSENCRYPT" == true ]; then
+    while [ -z "$FQDN" ]; do
+        echo -n "* Set the FQDN to use for Let's Encrypt (panel.example.com): "
```
Linux123123, line 590:
```suggestion
        echo -n "* Set the FQDN to use for Let's Encrypt (node.example.com): "
```

We should change this because people may be confused that it is panel and that they are doing a bad thing.

vilhelmprytz:
Script has now been tested to work. Thank you @Sam1370 for providing the testing instance!

Sam1370:
Great!

PR #98, opened 2020-09-30
panel: Reorder, fixes, certbot with --nginx

### LINKED ISSUE #88: Use Let's Encrypt --nginx over standalone webserver
The standalone webserver option could cause problems later on for renewals, since then nginx is running and occupying the HTTP/HTTPS ports.

Since certbot does support nginx it would be better to use `certbot --nginx`.

None.

**Additional context**
Mentioned in the Discord server.

Linux123123:
<b>Big reorder, added more dividers, made code simpler to read, changed up the order a bit, remade `perform_install`, remade OS specific installs, removed redundant `apt update`, removed unused variables, added `--nginx` instead of `--standalone`, removed a lot of redundant checks for OS version (if it checks if OS is the correct and that it is supported against a hard coded list, we don't need to recheck that every time we run something).</b>

<details>
<summary>Nothing here</summary>
<em>(I am 100% sure you are going to find stuff that you don't like that I added)</em>
</details>

<del><b>This is almost 99% not working because at the time of writing this PR, I haven't tested it.</b></del> 
<b> EDIT: working as of right now.</b>

Tested on:
Ubuntu 20.04
Debian 10
CentOS 7/8

Also there are going to be needed more changes to how we implement `nginx` config, because `certbot` itself is going to place all the needed certificate lines. I am leaving this untouched as of right now.

<b>Planning on doing the same on wings so wait for PR.</b>

A review is welcome :+1:   

PS: Don't hurry too much I have time.

Fixes #88


install-panel.sh
```diff
@@ -273,8 +343,56 @@ function ptdl_dl {
...
+    mysql -u root -e "FLUSH PRIVILEGES;"
+
+    echo "* MySQL database created & configured!"
+  fi
+}
+
+# Configure environment
 function configure {
-  [ "$ASSUME_SSL" == true ] && app_url=https://$FQDN || app_url=http://$FQDN
+  [ "$ASSUME_SSL" == true ] && app_url="https://$FQDN" || app_url="http://$FQDN"
```
vilhelmprytz, line 395: Why did you change this?

install-panel.sh
```diff
@@ -608,8 +671,7 @@ function centos_php {
 }
 
 function firewall_ufw {
-  apt update
-  apt install ufw -y
+  apt install -y ufw
```
vilhelmprytz, line 674: Haha, nice one... :)

install-panel.sh
```diff
@@ -665,21 +716,18 @@ function letsencrypt {
...
     print_error "OS not supported."
     exit 1
   fi
 
-  # Stop nginx
-  systemctl stop nginx
-
   # Obtain certificate
-  certbot certonly --no-eff-email --email "$email" --standalone -d "$FQDN" || FAILED=true
+  certbot certonly --no-eff-email --email "$email" --nginx -d "$FQDN" || FAILED=true
```
vilhelmprytz, line 730: Does this work with auto-renewal or does it have to spin up a temporary web server (therefore being blocked since nginx is running)?

vilhelmprytz (commented):
Thank you for the contribution, you're doing all my work now.... 🙂 I have only briefly reviewed it but I left some comments.

Linux123123:
So in reply to all the comments about `function`. Then we need to change all of them to have `function` or not have it. Also there are a few differences between them. You can find them [here](https://unix.stackexchange.com/a/73854).

install-panel.sh
```diff
@@ -117,11 +142,15 @@ function print_brake {
...
+function hyperlink() {
   echo -e "\e]8;;${1}\a${1}\e]8;;\a"
 }
 
-required_input() {
+#################################
+##### User input functions ######
+#################################
+
+function required_input() {
```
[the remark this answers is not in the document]
Linux123123: Then I would like to ask where else it should be defined? What the category should be called? I think I grouped them pretty good, because this section also has  `ask_firewall` and `ask_letsencrypt`.

install-panel.sh
Linux123123: Is there a big difference? I thought it looks nicer and helps in recognizing that this is a string defined.

install-panel.sh
Linux123123: Updated, should work now.

vilhelmprytz:
> So in reply to all the comments about `function`. Then we need to change all of them to have `function` or not have it. Also there are a few differences between them. You can find them [here](https://unix.stackexchange.com/a/73854).

Alright, I am a little bit unsure then. I know I had some trouble with global and local variables, perhaps defining the typeset would a better approach. I think the cleanest method should be used at all times. Perhaps this differs depending on if the function needs scoped variables or not?

Linux123123:
Well the choice is yours. I would say we ether use one or the other for consistency. I don't think that we will have any problems ether way.

install-panel.sh
Linux123123: Tested. Auto-renewal is working. Had to change quite a few things to get it working, but I got it working.
It doesn't spin up any temporary servers, because it is using `nginx` to do it all.

vilhelmprytz:
@Linux123123 Same as with #99. I did not close this on purpose, it was automatically closed since #52 was merged. Are you able to retain your changes and re-open a PR to `master` instead?

PR #100, opened 2020-10-12
panel: Reorder, fixes, certbot with --nginx

### LINKED ISSUE #88: Use Let's Encrypt --nginx over standalone webserver

Linux123123:
<b>Big reorder, added more dividers, made code simpler to read, changed up the order a bit, remade `perform_install`, remade OS specific installs, removed redundant `apt update`, removed unused variables, added `--nginx` instead of `--standalone`, removed a lot of redundant checks for OS version (if it checks if OS is the correct and that it is supported against a hard coded list, we don't need to recheck that every time we run something).</b>

|                | No SSL              | ASSUME SSL | SSL |
|----------------|---------------------|------------|-----|
| Ubuntu 20.04   | :white_check_mark: | :white_check_mark: | :white_check_mark: |
| Ubuntu 18.04   | :white_check_mark: | :white_check_mark: | :white_check_mark: |
| Debian buster  | :white_check_mark: | :white_check_mark: | :white_check_mark: |
| Debian stretch | :white_check_mark: | :white_check_mark: | :white_check_mark: |
| CentOS 7 *       | :white_check_mark: | :white_check_mark: | :white_check_mark: |
| CentOS 8 *       | :white_check_mark: | :white_check_mark: | :white_check_mark: |

Fixes #88


Linux123123:
@vilhelmprytz A thing to say is that when using `--nginx` when it fails it doesn't edit any files, so if it fails it just makes the site with no ssl.
Also in this PR you can see that there are no `nginx_ssl`. This is because option `--nginx` automatically adds all the needed thing for ssl and converts a no ssl config to ssl config.
So basically there is no way to `assume_ssl` after using certbot with  `--nginx`, because the files aren't edited if certificates fail. I don't see this as a flaw.

vilhelmprytz:
> @vilhelmprytz A thing to say is that when using `--nginx` when it fails it doesn't edit any files, so if it fails it just makes the site with no ssl.
> Also in this PR you can see that there are no `nginx_ssl`. This is because option `--nginx` automatically adds all the needed thing for ssl and converts a no ssl config to ssl config.
> So basically there is no way to `assume_ssl` after using certbot with `--nginx`, because the files aren't edited if certificates fail. I don't see this as a flaw.

Yes, I see your point. I remember I mentioned I was going to implement certbot `--nginx` myself separately? This PR is way too big for me to do in one go, you've piled up too much work in one PR. Is it too late to ask if you can split it up in smaller separate PRs?

Linux123123:
> > @vilhelmprytz A thing to say is that when using `--nginx` when it fails it doesn't edit any files, so if it fails it just makes the site with no ssl.
> > Also in this PR you can see that there are no `nginx_ssl`. This is because option `--nginx` automatically adds all the needed thing for ssl and converts a no ssl config to ssl config.
> > So basically there is no way to `assume_ssl` after using certbot with `--nginx`, because the files aren't edited if certificates fail. I don't see this as a flaw.
> 
> Yes, I see your point. I remember I mentioned I was going to implement certbot `--nginx` myself separately? This PR is way too big for me to do in one go, you've piled up too much work in one PR. Is it too late to ask if you can split it up in smaller separate PRs?

A bit late I would say. I have a solution. We can do the review threw discord and quickly go threw all of the stuff that I have done. It would be really hard for me to split this PR.

vilhelmprytz:
@Linux123123 This has conflicts that must be resolved before proper review :slightly_smiling_face: Let me know once you've solved the conflicts or if there is anything else I can look at in the meantime.

Linux123123:
This PR was started so long ago.... I completly forgot about it. Anyway I have merged and updated, so everything should be good.

install-panel.sh
```diff
@@ -1040,12 +938,11 @@ function goodbye {
...
-  echo "* Installation is using $WEBSERVER on $OS"
+  echo "* Installation is using nginx on $OS"
   echo "* Thank you for using this script."
   echo -e "* ${COLOR_RED}Note${COLOR_NC}: If you haven't configured the firewall: 80/443 (HTTP/HTTPS) is required to be open!"
   print_brake 62
 }
 
 # run script
-main
-goodbye
```
vilhelmprytz, line 1051: This is not consistent with the Wings script. Open a PR to edit the Wings script or modify the PR on this one. It doesn't really matter to me, not really a necessary change.

install-panel.sh
```diff
@@ -1040,12 +938,11 @@ function goodbye {
...
+main
```
vilhelmprytz, line 948: Are you able to keep the last new line?

vilhelmprytz (changes requested):

install-panel.sh
```diff
@@ -41,27 +45,19 @@ if ! [ -x "$(command -v curl)" ]; then
...
 fi
 
-# define version using information from GitHub
-get_latest_release() {
-  curl --silent "https://api.github.com/repos/$1/releases/latest" | # Get latest release from GitHub api
-  grep '"tag_name":' |                                              # Get tag line
-  sed -E 's/.*"([^"]+)".*/\1/'                                      # Pluck JSON value
-}
-
-echo "* Retrieving release information.."
```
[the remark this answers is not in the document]
Linux123123: Fixed.

install-panel.sh
```diff
@@ -585,139 +631,85 @@ function centos8_dep {
...
 
   echo -e "\n* Enabling Uncomplicated Firewall (UFW)"
-  echo "* Opening port 22 (SSH), 80 (HTTP) and 443 (HTTPS)"
+  echo "* Opening port 22 (SSH), 80 (HTTP) and (optionaly) 443 (HTTPS)"
 
   # pointing to /dev/null silences the command output
   ufw allow ssh > /dev/null
   ufw allow http > /dev/null
-  ufw allow https > /dev/null
+  [ "$CONFIGURE_LETSENCRYPT" == true ] && ufw allow https > /dev/null
```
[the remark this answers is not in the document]
Linux123123: I think that using a firewall, is only useful when you have only ports you need open. And if people want to enable https it's a pretty easy command.

Yeah, it should be `$ASSUME_SSL`

install-panel.sh
```diff
@@ -585,139 +631,85 @@ function centos8_dep {
...
+      ;;
+    centos)
+      [ "$OS_VER_MAJOR" == "7" ] && yum -y -q install certbot python-certbot-nginx
+      [ "$OS_VER_MAJOR" == "8" ] && dnf -y -q install certbot python3-certbot-nginx
+      ;;
+  esac
 
   # Obtain certificate
-  certbot certonly --no-eff-email --email "$email" --standalone -d "$FQDN" || FAILED=true
+  certbot --nginx --redirect --no-eff-email --email "$email" -d "$FQDN" || FAILED=true
```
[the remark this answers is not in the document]
Linux123123: So as I remember the last time I checked I am assigned to this. It doesn't change that much.

1. The nginx_ssl.conf is removed because using `--nginx` it creates ssl config from the non ssl config.
2. Assuming ssl isn't that straight forward. I left all the lines for getting user input, but it doesn't work.

Problem is talked about in:

> So basically there is no way to assume_ssl after using certbot with --nginx, because the files aren't edited if certificates fail. I don't see this as a flaw.

Now that I think I should add assuming ssl back, but in a bit different form. It would only be available in the start and if certificates fail, users would have the non-ssl config.
Or another option might be re downloading the nginx config files and editing them the second time, but with assume ssl.

install-panel.sh
```diff
@@ -744,118 +736,55 @@ function configure_nginx {
       ln -s /etc/nginx/sites-available/pterodactyl.conf /etc/nginx/sites-enabled/pterodactyl.conf
   fi
 
-  # restart nginx
-  if [ "$CONFIGURE_LETSENCRYPT" == true ] || { [ "$CONFIGURE_LETSENCRYPT" == false ] && [ "$ASSUME_SSL" == false ]; }; then
-    systemctl restart nginx
-  fi
```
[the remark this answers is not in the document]
Linux123123: The `--nginx` option manages everything by itself and this is not needed that much anymore. But by looking at it now it is needed when ssl is not used. Fixed.

install-panel.sh
Linux123123: Fixed.

install-panel.sh
Linux123123: Fixed. Probably caused by my editor.

install-panel.sh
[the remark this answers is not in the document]
vilhelmprytz: Sure, but 443 is really only going to have proper web servers listen to it and those would be vulnerable to any Denial of Service attacks on port 80 anyway.

install-panel.sh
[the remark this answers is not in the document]
vilhelmprytz: > So as I remember the last time I checked I am assigned to this. It doesn't change that much.
> 
> 1. The nginx_ssl.conf is removed because using `--nginx` it creates ssl config from the non ssl config.
> 2. Assuming ssl isn't that straight forward. I left all the lines for getting user input, but it doesn't work.
> 
> Problem is talked about in:
> 
> > So basically there is no way to assume_ssl after using certbot with --nginx, because the files aren't edited if certificates fail. I don't see this as a flaw.
> 
> Now that I think I should add assuming ssl back, but in a bit different form. It would only be available in the start and if certificates fail, users would have the non-ssl config.
> Or another option might be re downloading the nginx config files and editing them the second time, but with assume ssl.

I assigned it to you after you submitted the PR, not before. Not too sure what you mean here. If people want to assume SSL but not use Let's Encrypt, the `nginx_ssl.conf` file perhaps is best to be kept but not downloaded when using Nginx; only when using `assume_ssl`?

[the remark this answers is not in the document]
Linux123123: Yeah that's pretty easy, but the hard thing is assuming ssl after a failed letsencrypt.. It could be done, and I think I have a solution in my head.

I can easily add back assume ssl if user chooses out right, and at the same time think up the solution to assuming ssl after failed letsencrypt.

install-panel.sh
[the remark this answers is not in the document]
Linux123123: At the end it's your choice if we keep like this or change back. I think it's 50/50 for me.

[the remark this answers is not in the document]
vilhelmprytz: I say keep it like it was :slightly_smiling_face: I am open for further arguments against that decision though.

install-panel.sh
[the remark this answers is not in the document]
Linux123123: Ok I think I have added back assuming ssl.

install-panel.sh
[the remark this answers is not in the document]
Linux123123: Fixed.

install-panel.sh
```diff
@@ -744,118 +754,59 @@ function configure_nginx {
       ln -s /etc/nginx/sites-available/pterodactyl.conf /etc/nginx/sites-enabled/pterodactyl.conf
   fi
 
-  # restart nginx
-  if [ "$CONFIGURE_LETSENCRYPT" == true ] || { [ "$CONFIGURE_LETSENCRYPT" == false ] && [ "$ASSUME_SSL" == false ]; }; then
+  if [ ! "$ASSUME_SSL" ] && [ ! "$CONFIGURE_LETSENCRYPT" ]; then
     systemctl restart nginx
```
vilhelmprytz, line 758: Don't we want it to be restarted even when Certbot runs or is this handled somewhere else?

install-panel.sh
```diff
@@ -1040,12 +955,13 @@ function goodbye {
...
-  echo "* Installation is using $WEBSERVER on $OS"
+  echo "* Installation is using nginx on $OS"
   echo "* Thank you for using this script."
   echo -e "* ${COLOR_RED}Note${COLOR_NC}: If you haven't configured the firewall: 80/443 (HTTP/HTTPS) is required to be open!"
   print_brake 62
 }
 
 # run script
 main
+# Run goodbye function
```
vilhelmprytz, line 966: A bit redundant, is it not?

vilhelmprytz (commented):
Looking good. Have you tried the script on all supported installs?

install-panel.sh
Linux123123: If` certbot` runs we don't need to restart, because using `--nginx` already does that.

install-panel.sh
Linux123123: Yeah, probably.

Linux123123:
> Looking good. Have you tried the script on all supported installs?

I have tried on all supported installs about 3 months ago and they all worked.

I haven't tried with all of these changes that have been made in the last 3 months and don't really have a place test it right now.

install-panel.sh
```diff
@@ -690,23 +703,22 @@ function letsencrypt {
...
 #######################################
 ## WEBSERVER CONFIGURATION FUNCTIONS ##
 #######################################
 
-function configure_nginx {
+configure_nginx() {
   echo "* Configuring nginx .."
 
-  if [ "$ASSUME_SSL" == true ]; then
+  if "$ASSUME_SSL" && ! "$CONFIGURE_LETSENCRYPT"; then
```
vilhelmprytz, line 721: Why was the `[`, `]` removed?

install-panel.sh
```diff
@@ -744,118 +756,59 @@ function configure_nginx {
       ln -s /etc/nginx/sites-available/pterodactyl.conf /etc/nginx/sites-enabled/pterodactyl.conf
   fi
 
-  # restart nginx
-  if [ "$CONFIGURE_LETSENCRYPT" == true ] || { [ "$CONFIGURE_LETSENCRYPT" == false ] && [ "$ASSUME_SSL" == false ]; }; then
+  if "$ASSUME_SSL" && "$CONFIGURE_LETSENCRYPT"; then
```
vilhelmprytz, line 759: Same here about the square brackets. Also, doesn't this only restart if `ASSUME_SSL` and `CONFIGURE_LETSENCRYPT` are both true? For which cases do we need to restart Nginx? It should only be when `ASSUME_SSL` and `CONFIGURE_LETSENCRYPT` are both false (user is not using HTTPS) since you said certbot restarts Nginx automatically and when user is not using Let's Encrypt but they are planning on using SSL, they need to modify their config file before restarting Nginx manually. Am I missing something obvious here?

install-panel.sh
Linux123123: Because with them it didn't work right.

install-panel.sh
Linux123123: They can't be both true.

install-panel.sh
vilhelmprytz: I am a little unsure of what consequences not using square brackets has when it comes to functionality. As far as I know, the square bracket in bash is a shorthand notation for performing a conditional test. When using square brackets, bash "tests" the expression. By not using them, you are essentially testing the variable as an expression directly. In this case, `ASSUME_SSL` and `CONFIGURE_LETSENCRYPT` should not be anything else than booleans which leads to the same result since `true` has exit-code `0` when used as an expression and `false` has exit-code `1` when used as an expression.

install-panel.sh
vilhelmprytz: Which variables cannot be true simultaneously?

Linux123123: `ASSUME_SSL` and `CONFIGURE_LETSENCRYPT` can't be true at the same time.

install-panel.sh
Linux123123: Ok, I will test once again.

install-panel.sh
vilhelmprytz: Can they not? I think in the `ask_letsencrypt` function they are both set to true. See https://github.com/Linux123123/pterodactyl-installer/blob/95a1f89bf90347fc936d1375bde7be4d32d49d41/install-panel.sh#L196

Linux123123: Yes, you are right. It's an issue. It was left behind, because I changed the way how assuming ssl works. There it should only be set `CONFIGURE_LETSENCRYPT`.

install-panel.sh
Linux123123: Ok I played around a bit,
```bash
#!/bin/bash

i=false
k=true

if $i && $k; then
  echo "Both true"
elif $i && ! $k; then
  echo "k not true"
elif ! $i && $k; then
  echo "i not true"
fi
```

And it works without any problem.

Linux123123: Do you have a better solution (that involves `[ ]` )? I couldn't get it to work with `[ ]`. Also be sure that it is compliant with shellcheck.

install-panel.sh
vilhelmprytz: So what purpose do the variables have now? `ASSUME_SSL` should only be true if `CONFIGURE_LETSENCRYPT` is false and the user then wants to use their own certificate. If `CONFIGURE_LETSENCRYPT` is true the user wants to use Let's Encrypt automatically but the `ASSUME_SSL` should **not** be true. And if both are false, just pure HTTP. Am I correct?

install-panel.sh
vilhelmprytz: I've updated your PR. You think it looks OK? So using `!` basically means "not" whereas the variable is a boolean, true or false. Also, you were using `"` which actually makes the booleans into "strings" so that may have caused problems :slightly_smiling_face: You can rewrite whatever I wrote now but a bit more compact but this works as well!

install-panel.sh
Linux123123: Yes, you are correct.

vilhelmprytz: Alright, considered resolved then.

install-panel.sh
```diff
@@ -28,9 +28,9 @@ set -e
 #                                                                           #
 #############################################################################
 
-# versioning
-GITHUB_SOURCE="master"
-SCRIPT_RELEASE="canary"
+#################################
+######## General checks #########
+#################################
```
vilhelmprytz, line 33: Are these large comments required? They add three new lines per comment which is quite a lot.

install-panel.sh
```diff
@@ -45,27 +45,23 @@ if ! [ -x "$(command -v curl)" ]; then
...
-echo "* Retrieving release information.."
-PTERODACTYL_VERSION="$(get_latest_release "pterodactyl/panel")"
-echo "* pterodactyl/panel @ $PTERODACTYL_VERSION"
+# versioning
+GITHUB_SOURCE="master"
+SCRIPT_RELEASE="canary"
 
-# variables
-WEBSERVER="nginx"
+# No default FQDN
```
vilhelmprytz, line 56: Why is it important that the FQDN variable does not have a default value?

install-panel.sh
```diff
@@ -209,37 +266,27 @@ function detect_distro {
...
-    elif [ "$OS_VER_MAJOR" == "8" ]; then
-      SUPPORTED=true
-    else
-      SUPPORTED=false
-    fi
-  else
-    SUPPORTED=false
-  fi
+check_os_comp() {
+  SUPPORTED=false
```
vilhelmprytz, line 270: Is this redundant? The default case, handled with the wildcard, also sets `SUPPORTED=false`. I have not tested if this is the case or not.

install-panel.sh
```diff
@@ -277,8 +326,56 @@ function ptdl_dl {
...
+    echo "* Flushing privileges.."
+    mysql -u root -e "FLUSH PRIVILEGES;"
+
+    echo "* MySQL database created & configured!"
+  fi
+}
+
+# Configure environment
+configure() {
+  [ "$ASSUME_SSL" == true ] && app_url="https://$FQDN" || app_url="http://$FQDN"
```
vilhelmprytz, line 378: For some reason, this does not work with zsh. Works fine with bash though so no problems.

This does not reflect the case where `CONFIGURE_LETSENCRYPT` is true. I am a bit unsure whether it was necessary to change the behavior of `ASSUME_SSL` and `CONFIGURE_LETSENCRYPT`. Either way, this must be changed to reflect the new variable standards.

install-panel.sh
```diff
@@ -354,109 +444,90 @@ function install_pteroq {
...
+  systemctl enable redis
+  systemctl start mariadb
+  systemctl start nginx
+  systemctl start redis
+}
 
-function apt_update {
-  apt update -y && apt upgrade -y
+selinux_allow() {
+  setsebool -P httpd_can_network_connect 1 || true
```
vilhelmprytz, line 480:
```suggestion
  setsebool -P httpd_can_network_connect 1 || true  # these commands can fail OK
```

install-panel.sh
```diff
@@ -354,109 +444,90 @@ function install_pteroq {
...
+  systemctl start mariadb
+  systemctl start redis-server
 }
 
-##################################
-# OS specific install functions ##
-##################################
+enable_services_centos_based() {
+  systemctl enable mariadb
+  systemctl enable nginx
```
vilhelmprytz, line 472: Do we have to enable nginx on CentOS?

install-panel.sh
```diff
@@ -589,34 +638,18 @@ function centos8_dep {
 ## OTHER OS SPECIFIC FUNCTIONS ##
 #################################
 
-function ubuntu_universedep {
```
vilhelmprytz, line 592: Thank you for removing this. Why did I even implement it like this in the first place? Makes no sense even to me :slightly_smiling_face:

install-panel.sh
```diff
@@ -628,63 +661,43 @@ function firewall_ufw {
...
-    echo "* Installing firewall"
-    dnf -y -q update > /dev/null
-    dnf -y -q install firewalld > /dev/null
-
-    systemctl --now enable firewalld > /dev/null # Start and enable
-    firewall-cmd --add-service=http --permanent -q # Port 80
-    firewall-cmd --add-service=https --permanent -q # Port 443
-    firewall-cmd --add-service=ssh --permanent -q  # Port 22
-    firewall-cmd --reload -q # Enable firewall
+  echo "* Opening port 22 (SSH), 80 (HTTP) and (optionaly) 443 (HTTPS)"
```
vilhelmprytz, line 666:
```suggestion
  echo "* Opening port 22 (SSH), 80 (HTTP) and 443 (HTTPS)"
```

install-panel.sh
```diff
@@ -589,34 +638,18 @@ function centos8_dep {
...
 
-function firewall_ufw {
-  apt update
-  apt install ufw -y
+firewall_ufw() {
+  apt install -y ufw
 
   echo -e "\n* Enabling Uncomplicated Firewall (UFW)"
-  echo "* Opening port 22 (SSH), 80 (HTTP) and 443 (HTTPS)"
+  echo "* Opening port 22 (SSH), 80 (HTTP) and (optionaly) 443 (HTTPS)"
```
vilhelmprytz, line 652:
```suggestion
  echo "* Opening port 22 (SSH), 80 (HTTP) and 443 (HTTPS)"
```

The word `optionaly` should be spelled `optionally` but I think we can remove it all together :slightly_smiling_face:

install-panel.sh
```diff
@@ -628,63 +661,43 @@ function firewall_ufw {
...
-    # exit
-    print_error "OS not supported."
-    exit 1
-  fi
-
-  # Stop nginx
-  systemctl stop nginx
+  case "$OS" in
+    debian | ubuntu)
+      apt-get -y install certbot python3-certbot-nginx
```
vilhelmprytz, line 691: Is the `python3-certbot-nginx` required? Can't recall having to install it on Debian 10.

install-panel.sh
```diff
@@ -695,23 +708,22 @@ function letsencrypt {
     if [[ "$CONFIGURE_SSL" =~ [Yy] ]]; then
       ASSUME_SSL=true
       CONFIGURE_LETSENCRYPT=false
+      configure_nginx
```
vilhelmprytz, line 711: Just to be sure that you have handled all cases here (it looks to me like you have).  `configure_nginx` must be run when the user is using HTTP or just `ASSUME_SSL`. `configure_nginx` and then `letsencrypt` must be run when the user is using Let's Encrypt. And if the `letsencrypt` fails, and the user still opts in to use the `ASSUME_SSL`, then we need to re-run `configure_nginx`. Are all these cases handled? Does the `configure_nginx` function handle being re-run without machine wipe?

install-panel.sh
vilhelmprytz: Nevermind, still about the brackets.

install-panel.sh
```diff
@@ -749,118 +761,59 @@ function configure_nginx {
...
+  [ "$OS" == "centos" ] && centos_php
+  install_composer
+  ptdl_dl
+  create_database
+  configure
+  set_folder_permissions
+  insert_cronjob
+  install_pteroq
+  configure_nginx
+  "$CONFIGURE_LETSENCRYPT" && letsencrypt
```
vilhelmprytz, line 813: Also about the brackets, but perhaps it's okay if it works. `CONFIGURE_LETSENCRYPT` here is going to be treated as a string and not as a boolean.

vilhelmprytz:
Also sorry for taking so long before properly reviewing your PR. When you move a lot of the functions around in conjunction with all the changes, it makes it harder for me to review in one go.

I think we're closing in on a merge now. I've got to test the script using Let's Encrypt before.

install-panel.sh
Linux123123: But it is a lot easier to find the section you need with them. If you want me to remove them, I can.

install-panel.sh
Linux123123: Not really that important :)

install-panel.sh
Linux123123: Yes, I thought about removing and then forgot. Thanks for reminding.

install-panel.sh
Linux123123: Ok, will push a commit to fix this.

install-panel.sh
Linux123123: Yes, I just tested, `dnf` or `yum` doesn't enable or start `nginx`

install-panel.sh
Linux123123: Hehe....

install-panel.sh
Linux123123:
```bash
root@localhost:~# apt install certbot
Reading package lists... Done
Building dependency tree
Reading state information... Done
The following additional packages will be installed:
  python3-acme python3-certbot python3-configargparse python3-josepy python3-mock
  python3-openssl python3-parsedatetime python3-pbr python3-requests-toolbelt
  python3-rfc3339 python3-tz python3-zope.component python3-zope.event
  python3-zope.hookable python3-zope.interface
Suggested packages:
  python3-certbot-apache python3-certbot-nginx python-certbot-doc python-acme-doc
  python-mock-doc python-openssl-doc python3-openssl-dbg
Recommended packages:
  python3-pyicu
The following NEW packages will be installed:
  certbot python3-acme python3-certbot python3-configargparse python3-josepy
  python3-mock python3-openssl python3-parsedatetime python3-pbr
  python3-requests-toolbelt python3-rfc3339 python3-tz python3-zope.component
  python3-zope.event python3-zope.hookable python3-zope.interface
```
This is debian stretch (shouldn't be different from buster). It suggests installing `python3-certbot-nginx `, but doesn't install it. It is needed for `certbot --nginx`

install-panel.sh
Linux123123: It handles that alright. Just `curl` overwrites the files, `ln` links the same thing, `sed` replaces variables as before and `rm -rf` doesn't delete anything.

install-panel.sh
Linux123123: So all good?

install-panel.sh
Linux123123: Fixed.

install-panel.sh
Linux123123: Fixed.

install-panel.sh
Linux123123: It might have problems, because I need to reboot machine to have a working panel at least on CentOS 8 (also maybe on CentOS 7).

Linux123123: Ok enabling might be the only thing needed for CentOS. I didn't get that pid error this time, so maybe it has been fixed.

Linux123123:
Ok all OS's are tested now.

install-panel.sh
vilhelmprytz: I think they should be removed.

Linux123123: And replaced by anything?

vilhelmprytz: One line comments.

Linux123123: So from 
```bash
######## General checks #########
```
to
```bash
######## General checks #########
```
?

install-panel.sh
```diff
@@ -171,8 +175,58 @@ password_input() {
...
+}
+
+ask_assume_ssl() {
+  echo "* Let's Encrypt is not going to be automatically configured by this script (user opted out)."
+  echo "* You can 'assume' Let's Encrypt, which means the script will download a nginx configuration that is configured to use a Let's Encrypt certificate but the script won't obtain the certificate for you."
+  echo "* If you assume SSL and do not obtain the certificate, your installation will not work."
+  echo -n "* Assume SSL or not? (y/N): "
+  read -r ASSUME_SSL_INPUT
+
+  [[ "$ASSUME_SSL_INPUT" =~ [Yy] ]] && ASSUME_SSL=true
```
vilhelmprytz, line 201:
```suggestion
  [[ "$ASSUME_SSL_INPUT" =~ [Yy] ]] && ASSUME_SSL=true || true
```

install-panel.sh
```diff
@@ -749,118 +752,57 @@ function configure_nginx {
...
+  [ "$OS" == "centos" ] && centos_php
+  install_composer
+  ptdl_dl
+  create_database
+  configure
+  set_folder_permissions
+  insert_cronjob
+  install_pteroq
+  configure_nginx
+  [ "$CONFIGURE_LETSENCRYPT" == true ] && letsencrypt
```
vilhelmprytz, line 802:
```suggestion
  [ "$CONFIGURE_LETSENCRYPT" == true ] && letsencrypt
  true
```

Since `[ "$CONFIGURE_LETSENCRYPT" == true ] && letsencrypt` is the last command in the body of the function `main`, it's exit code defines how `main` exits. Therefore, not using Let's Encrypt would mean that the `goodbye` function never runs since the script would exit here.

vilhelmprytz (approved):
LGTM :+1:

PR #101, opened 2020-10-12
wings: Reorder, remove unneded stuff

Linux123123:
<b>Big reorder, added more dividers, made code simpler to read, changed up the order a bit, removed redundant apt update, removed a lot of redundant checks for OS version (if it checks if OS is the correct and that it is supported against a hard coded list, we don't need to recheck that every time we run something).</b>

<del>This is almost 99% not working because at the time of writing this PR, I haven't tested it.</del>
<b> EDIT: working as of right now.</b>

Tested on:
Ubuntu 20.04
Debian 10
CentOS 7/8

PS: Don't hurry too much I have time.


install-wings.sh
```diff
@@ -447,70 +390,85 @@ function firewall_ufw {
...
-  if [[ "$CONFIRM_SSL" =~ [Yy] ]]; then
-    CONFIGURE_LETSENCRYPT=true
-  fi
-}
-
-function main {
+main() {
   # check if we can detect an already existing installation
-  if [ -d "/etc/pterodactyl" ]; then
+  if [ -d "/etc/pterodactyl/" ]; then
```
vilhelmprytz, line 471: Not too sure the reasoning behind this change?

Linux123123: hmmmn copying mistake?

vilhelmprytz: Sorry, what do you mean?

Linux123123: So I should remove it? It  doesn't make any difference after all.

vilhelmprytz: Well yes, it isn't more correct to do this way what?

Linux123123: In my Linux experience I felt that having a `/` at the end made it look more as directory and not a file. It's just personal experience so I will change it. :+1:

install-wings.sh
```diff
@@ -153,33 +169,22 @@ function check_os_comp {
...
+    debian)
+      [ "$OS_VER_MAJOR" == "9" ] && SUPPORTED=true
+      [ "$OS_VER_MAJOR" == "10" ] && SUPPORTED=true
+      ;;
+    centos)
+      [ "$OS_VER_MAJOR" == "7" ] && SUPPORTED=true
+      [ "$OS_VER_MAJOR" == "8" ] && SUPPORTED=true
+      ;;
+    *)
+      SUPPORTED=false ;;
```
vilhelmprytz, line 186: Nice, this looks a lot cleaner! :+1:

install-wings.sh
```diff
@@ -220,15 +225,19 @@ function check_os_comp {
...
-    exit 1
-  fi
+  case "$virt_serv" in
+    openvz | lxc)
+      print_warning "Unsupported type of virtualization detected. Please consult with your hosting provider whether your server can run Docker or not. Proceed at your own risk."
+      print_error "Installation aborted!"
+      exit 1
+      ;;
+    "")
+      ;;
```
vilhelmprytz, line 236: Really nit-picking here, so disregard this as you like (since this PR is amazing as it is and this is by far better than it already is), but can you do this without catching the empty case? I understand that if `virt_serv` is empty it shouldn't do anything at all so this statement looks redundant to me? Perhaps I am missing something?

install-wings.sh
```diff
@@ -220,15 +225,19 @@ function check_os_comp {
...
+    *)
+      print_warning "Virtualization: ${virt_serv//$'\n'/ } detected."
```
vilhelmprytz, line 238: What's up with the `\n` at the end? I think I recall `virt_serv` output being extremely long.

install-wings.sh
```diff
@@ -240,157 +249,79 @@ function check_os_comp {
...
-    # add APT repo
-    sudo add-apt-repository \
-     "deb [arch=amd64] https://download.docker.com/linux/ubuntu \
-     $(lsb_release -cs) \
-     stable"
+    # Add docker repo
+    add-apt-repository \
+      "deb [arch=amd64] https://download.docker.com/linux/$OS \
+      $(lsb_release -cs) \
+      stable"
```
vilhelmprytz, line 286:
```suggestion
    "deb [arch=amd64] https://download.docker.com/linux/$OS \
    $(lsb_release -cs) \
    stable"
```

Shouldn't the indenting be like this?

install-wings.sh
```diff
@@ -428,8 +361,22 @@ function install_mariadb {
 ##### OS SPECIFIC FUNCTIONS #####
 #################################
 
-function firewall_ufw {
-  apt update
```
vilhelmprytz, line 432: Shouldn't the `apt update` be retained?

install-wings.sh
```diff
@@ -447,68 +394,83 @@ function firewall_ufw {
...
-  [ "$OS_VER_MAJOR" == "7" ] && yum -y -q update
   [ "$OS_VER_MAJOR" == "7" ] && yum -y -q install firewalld > /dev/null
-  [ "$OS_VER_MAJOR" == "8" ] && dnf -y -q update
   [ "$OS_VER_MAJOR" == "8" ] && dnf -y -q install firewalld > /dev/null
 
   # Enable
   systemctl --now enable firewalld > /dev/null # Enable and start
 
   # Configure
+  firewall-cmd --ad-service=ssh --permanent -q # Port 22
```
vilhelmprytz, line 409: Oops, this is a misspelling. I can see that you are not the one to blame since the invalid command is in the `master` branch as of right now! Has it been broken all this time? 😱
```suggestion
  firewall-cmd --add-service=ssh --permanent -q # Port 22
```

install-wings.sh
```diff
@@ -447,68 +394,83 @@ function firewall_ufw {
...
+perform_install() {
   echo "* Installing pterodactyl wings.."
-  [ "$CONFIGURE_UFW" == true ] && firewall_ufw
-  [ "$CONFIGURE_FIREWALL_CMD" == true ] && firewall_firewalld
-  install_dep
+  [ "$OS" == "ubuntu" ] || [ "$OS" == "debian" ] && apt_update
+  [ "$OS" == "centos" ] || [ "$OS_VER_MAJOR" == "7" ] && yum_update
+  [ "$OS" == "centos" ] || [ "$OS_VER_MAJOR" == "8" ] && dnf_update
+  "$CONFIGURE_UFW" && firewall_ufw
+  "$CONFIGURE_FIREWALL_CMD" && firewall_firewalld
```
vilhelmprytz, line 462: Nice refactor, a lot cleaner than checking for the value specifically! :+1:

install-wings.sh
```diff
@@ -535,8 +494,12 @@ function main {
   echo "* This script is not associated with the official Pterodactyl Project."
   echo "*"
   echo "* Running $OS version $OS_VER."
+  echo "* Latest pterodactyl wings version is $WINGS_VERSION"
   print_brake 70
 
+  # checks if the system is compatible with this installation script
+  check_os_comp
+
```
vilhelmprytz, line 502: Why did you move this?

install-wings.sh
```diff
@@ -586,10 +535,10 @@ function main {
     if [[ "$CONFIRM_FIREWALL_CMD" =~ [Yy] ]]; then
       CONFIGURE_FIREWALL_CMD=true
     fi
-
-    ask_letsencrypt
   fi
 
+  ask_letsencrypt
```
vilhelmprytz, line 540: This means all supported distros will at this stage try to install Let's Encrypt if the user opts in to do so. Have you verified that all supported distros and distro releases work with the Let's Encrypt the way we intended it?

vilhelmprytz (commented):
Fantastic work, overall I think this looks amazing! Since you've already tested it and everything I reckon this is quite near merging! :+1:

install-wings.sh
Linux123123: Thanks :+1:

install-wings.sh
Linux123123: Thanks :+1:

install-wings.sh
Linux123123: image

It doesn't show me that like that, but I will commit yours.

install-wings.sh
Linux123123: No, because It updates one time only at the start and as far as I know` ufw` comes after that

image

install-wings.sh
Linux123123: So that it tells what OS it sees and only then stops the script. Then user can see what OS is running. Of course it also show with print_errors. Now I don't know what to say.

install-wings.sh
Linux123123: As far as I know Debian 9,10; Ubuntu 18.04, 20.04; CentOS 7, 8 works.

I think that's all the supported OS's.

install-wings.sh
Linux123123: So this is here, because it should give an error when the `virt_serv` string is empty. I actually searched for a better solution, but apart from regex, couldn't find anything. Going back to `if` might also be an option.

install-wings.sh
Linux123123: I think this doesn't make it go to a new line. (it removes \n from end). I copied this from internet on my first PR.

install-wings.sh
vilhelmprytz: Sorry, I am not too sure what you mean. `virt_serv` is empty when there is no virtualization detected, which is quite common for a lot of people running Wings. Right now, the code will not print anything when `virt_serv` is empty but it will print the warning below if it's not empty nor handled by the case above. If you think it should give an error when `virt_serv` is empty, that is not what the code is doing currently.

install-wings.sh
vilhelmprytz:
```bash
$ export test="test"
$ echo "${test//$'\n'/ }"
test
$ export test="test\n\n\n"
$ echo "${test//$'\n'/ }"
test



``` 

Am I doing something wrong here? It does not remove the `\n`' for me.

install-wings.sh
vilhelmprytz: I guess it's just as good if it's here. I'll leave it.

install-wings.sh
vilhelmprytz: Have you tried the Let's Encrypt function on the supported distros? I am not questioning the fact that you are probably right, when we supported Ubuntu 16 this was not the case, hence the version compatibility checking was implemented in the first place.

vilhelmprytz:
The branch is behind upstream, consider pulling the new changes (if possible) :+1:

install-wings.sh
Linux123123: When I did the testing I did it on all OS's above. That's why I said they all work. When I test I also test ssl just in case I did something wrong there.

install-wings.sh
Linux123123: Ok first of all I had a typo there and instead of 'shouldn't' I wrote 'should'. 

Next thing is that there isn't any other solution when using `case`, because it needs to print warning when `lxc` or `openvz` and it needs to print warning when that string is not empty. If I remove that `"")` it will print the warning without any virtualization inside of it. Of course a simple check for an empty string at the start of `print_warning`  would be viable. Or as I said before a fallback to `if` is also viable.

install-wings.sh
Linux123123: Ok tested on my system and seems that it doesn't do anything too. Will revert back normal.

install-wings.sh
vilhelmprytz: It looks to me like you solved it in another way! :+1:

Linux123123: Yeah just was tired of overthinking this.

install-wings.sh
vilhelmprytz: Alright, all cool!

Linux123123: Cool. cool.

install-wings.sh
```diff
@@ -41,17 +45,9 @@ if ! [ -x "$(command -v curl)" ]; then
...
 fi
 
-# define version using information from GitHub
-get_latest_release() {
-  curl --silent "https://api.github.com/repos/$1/releases/latest" | # Get latest release from GitHub api
-    grep '"tag_name":' |                                            # Get tag line
-    sed -E 's/.*"([^"]+)".*/\1/'                                    # Pluck JSON value
-}
-
-echo "* Retrieving release information.."
```
vilhelmprytz, line 51: This message is not re-added down below along with the `get_latest_release` function. Do you think it is unnecessary to retain the info message? Retrieving information about the latest release can sometimes take a few seconds, especially when the GitHub API is unresponsive which is why it was there in the first place. But I am open to changes, of course, I'd just like to hear your reasoning behind the change and your point of view.

install-wings.sh
```diff
@@ -240,157 +247,79 @@ function check_os_comp {
...
+      gnupg2 \
       software-properties-common
 
-    # get their GPG key
-    curl -fsSL https://download.docker.com/linux/ubuntu/gpg | sudo apt-key add -
+    # Add docker gpg key
+    curl -fsSL https://download.docker.com/linux/"$OS"/gpg | apt-key add -
 
-    # show fingerprint to user
-    apt-key fingerprint 0EBFCD88
```
vilhelmprytz, line 346: I'd like the fingerprint to still be shown to the user. Most non-advanced users won't notice if anything is out of the ordinary, but if it is something bad with the GPG key, it will most likely show here.

vilhelmprytz (changes requested):
This is my final review. A few minor comments. We're getting close to merging :sunglasses:

install-wings.sh
Linux123123: I think we can remove the `get_latest_version` entirely, but that's just my opinion. I think if we show latest instead of version number it would be just as good. So should I add it back or remove?

install-wings.sh
Linux123123: Added back, done. :+1:

install-wings.sh
vilhelmprytz: I like the fact that it displays the version, and not just `latest` because it is so vague. I reckon you should add the `* Retrieving release information..` message back if the version fetching is going to be kept, which I think it should, does it sound reasonable?

Linux123123: Ok it's back...

Linux123123 (commented):
LGTM (at least to me) (also try installing at least once)

vilhelmprytz (approved):
LGTM 👍. Thank you very much for the contribution, this greatly improves general readability of the script!

PR #102, opened 2020-10-18
Upgrade scripts

### LINKED ISSUE #94: wings/panel: Option to update to latest versions
I think it would be really beneficial for the community to have an option on the master script that will update the panel and wings to the latest versions, so an update option.

vilhelmprytz:
Fixes #94


Linux123123:
Only wings? I thought the idea was to implement both? or did you only do wings because of time constraints?

vilhelmprytz:
> Only wings? I thought the idea was to implement both? or did you only do wings because of time constraints?

@Linux123123 This PR is still a draft. Panel is a bit more complicated since `php` has to be upgraded manually.

vilhelmprytz:
This PR is no longer active. I don't have the time to work on this right now, unfortunately.

PR #110, opened 2020-11-04
panel: setsebool commands on CentOS can now fail without abort

### LINKED ISSUE #106: Panel: SElinux unsupported on some providers
**Describe the bug**
The script won't run past the SElinux part, because it fails at that point.

**To Reproduce**
1. Run the script on CentOS 7 or 8
4. Come to the part where SElinux is enabled and get an error

**Expected behaviour**
The script should skip the SElinux part when it is unsupported.

 - OS: CentOS 7 and 8
 - Installation: panel

**Additional context**
This isn't a crucial install step, so it could be fixed with ` || true` or by checking if SElinux is supported by the kernel.

vilhelmprytz:
If the `setsebool` commands fail they will no longer cause the script to abort

Fixes #106

### DIFF
--- a/install-panel.sh
+++ b/install-panel.sh
@@ -538,9 +538,9 @@ function centos7_dep {
   systemctl start redis
 
   # SELinux (allow nginx and redis)
-  setsebool -P httpd_can_network_connect 1
-  setsebool -P httpd_execmem 1
-  setsebool -P httpd_unified 1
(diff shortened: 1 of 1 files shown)


vilhelmprytz:
@Linux123123 Is this what you were looking for?

Linux123123:
Yes I think.

PR #111, opened 2020-11-07
Add a way to verify that a FQDN is valid

vilhelmprytz:
Verifies that the FQDN provided by user is valid.

### DIFF
--- a/install-panel.sh
+++ b/install-panel.sh
@@ -938,7 +938,7 @@ function main {
   done
 
   # verify FQDN
-  bash <(curl -s $GITHUB_BASE_URL/lib/verify-fqdn.sh) "$FQDN"
+  bash <(curl -s $GITHUB_BASE_URL/lib/verify-fqdn.sh) "$FQDN" "$OS"
 
   # UFW is available for Ubuntu/Debian
   # Let's Encrypt is available for Ubuntu/Debian
--- a/install-panel.sh
+++ b/install-panel.sh
@@ -77,7 +77,7 @@ CONFIGURE_LETSENCRYPT=false
 
 # download URLs
 PANEL_DL_URL="https://github.com/pterodactyl/panel/releases/latest/download/panel.tar.gz"
-CONFIGS_URL="https://raw.githubusercontent.com/vilhelmprytz/pterodactyl-installer/master/configs"
+GITHUB_BASE_URL="https://raw.githubusercontent.com/vilhelmprytz/pterodactyl-installer/master"
 
 # apt sources path
 SOURCES_PATH="/etc/apt/sources.list"
@@ -343,7 +343,7 @@ function insert_cronjob {
 function install_pteroq {
   echo "* Installing pteroq service.."
 
-  curl -o /etc/systemd/system/pteroq.service $CONFIGS_URL/pteroq.service
+  curl -o /etc/systemd/system/pteroq.service $GITHUB_BASE_URL/configs/pteroq.service
   systemctl enable pteroq.service
   systemctl start pteroq
 
@@ -601,7 +601,7 @@ function ubuntu_universedep {
 }
 
 function centos_php {
-  curl -o /etc/php-fpm.d/www-pterodactyl.conf $CONFIGS_URL/www-pterodactyl.conf
+  curl -o /etc/php-fpm.d/www-pterodactyl.conf $GITHUB_BASE_URL/configs/www-pterodactyl.conf
 
   systemctl enable php-fpm
   systemctl start php-fpm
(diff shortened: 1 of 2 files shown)


MrFlacko:
I do have this code I did a while ago if any of this helps you out.

`PublicIP="$(wget http://ipecho.net/plain -O - -q ; echo)"

DomainTester() {
  read -p "Please enter the Fully Qualified Domain Name to be used (EXAMPLE: panel.testserver.com): " FQDN
  DomainIP="$(dig +short $FQDN)"
  [[ "$DomainIP" == "$PublicIP" ]] && return
  echo -e "\n${LIGHT_RED}IMPORTANT${NoColor}"
  echo -e "The Domain IP you entered $DomainIP isn't the same as your public ip $PublicIP"
  echo -e "This script may not work, make sure if you are using cloudflare you are running"
  echo -e "in dns-only mode or you know what you're doing."
  echo -e "Either way the SSL section of the script will fail, then starting NGINX will produce an error"
  echo
  read -p "Ctrl+C to stop the scipt, or you can continue by pressing [Enter]"
  clear
}

DomainTester`

vilhelmprytz:
> I do have this code I did a while ago if any of this helps you out.
> 
> ```
> PublicIP="$(wget http://ipecho.net/plain -O - -q ; echo)"
> 
> DomainTester() {
>   read -p "Please enter the Fully Qualified Domain Name to be used (EXAMPLE: panel.testserver.com): " FQDN
>   DomainIP="$(dig +short $FQDN)"
>   [[ "$DomainIP" == "$PublicIP" ]] && return
>   echo -e "\n${LIGHT_RED}IMPORTANT${NoColor}"
>   echo -e "The Domain IP you entered $DomainIP isn't the same as your public ip $PublicIP"
>   echo -e "This script may not work, make sure if you are using cloudflare you are running"
>   echo -e "in dns-only mode or you know what you're doing."
>   echo -e "Either way the SSL section of the script will fail, then starting NGINX will produce an error"
>   echo
>   read -p "Ctrl+C to stop the scipt, or you can continue by pressing [Enter]"
>   clear
> }
> 
> DomainTester
> ```

@MrFlacko Thank you for the contribution. I want to avoid using any third-party services (e.g. `ipecho.net` that you use in your example). Also, this would fail if the `FQDN` points to the server but a firewall outside the server is blocking the port.

MrFlacko:
Well the port it uses should be the same port as a domain would use. Just sniffing out port 53 (i think thats dns port)

but no worries, thanks for your feedback

vilhelmprytz:
> ain would use. Just sniffi

No, what? If you lookup the DNS-records for a domain name (or subdomain) you're device is going to ask the nameservers for the domain; it

MrFlacko:
Maybe I'm being dumb here, I though did used the same port to find the address, as it is to access the address through the domain. If the domain is publicly accessible then dig should be able to get the IP address. There shouldn't be a way to block that unless your domain isn't publicly accessible.

vilhelmprytz:
> Maybe I'm being dumb here, I though did used the same port to find the address, as it is to access the address through the domain. If the domain is publicly accessible then dig should be able to get the IP address. There shouldn't be a way to block that unless your domain isn't publicly accessible.

@MrFlacko I am not too sure what you mean. Let me explain my thought process with an example. Let's say my hostname is `vilhelmprytz.se` (that's my website). When I query using `dig +short vilhelmprytz.se`, I get the IP that is hosting my website. When I executed the DNS query, i.e. the `dig` commad, my computer spoke to the nameservers of the domain `vilhelmprytz.se`. You can lookup the nameservers using `dig +short -t ns vilhelmprytz.se`, or just a plain `whois vilhelmprytz.se`.

That gives me the following.

```
ns1.1984hosting.com.
ns0.1984.is.
ns2.1984hosting.com.
ns2.1984.is.
ns1.1984.is.
```

So when my computer queried `vilhelmprytz.se` for DNS-records using `dig`, my computer connected to one of these name servers (at least in practice, in reality, my ISP probably caches most DNS requests before my computer even sees them, but the end result will be the same). Never did my computer actually connect to the IP that the domain `vilhelmprytz.se` is pointing to, only to the nameservers responsible for the domain.

If I were to visit my website, [https://vilhelmprytz.se](https://vilhelmprytz.se), my computer would lookup the DNS-records and then connect to the IP of the server that the DNS servers return. This would fail if port 443 is closed. But when only querying with `dig` I am never speaking to the server itself therefore not validating that any ports are open.

Hope my example is clear enough.

JustinPooters:
i recall this script checking the FQDN 
https://github.com/Revenact/Pterodactyl-Script/
Maybe you could check it out and try to use it. ^-^

vilhelmprytz:
> i recall this script checking the FQDN
> https://github.com/Revenact/Pterodactyl-Script/
> Maybe you could check it out and try to use it. ^-^

@JustinPooters Thank you for the suggestion!

PR #114, opened 2020-11-13
Remove unnecessary firewall instructions from README

### LINKED ISSUE #112: README.md: firewall instructions not needed
**Describe the bug**
When you are first time on the page you might think that configuring firewall manually is a must, but it's not.

**To Reproduce**
1. Go to the github package
2. See firewall options
3. Think that they are a must

**Expected behaviour**
Users should use the automatic firewall install!!!!

**Screenshots**
image

 - OS: GNU/Linux Arch Linux
 - Installation: panel and wings

**Additional context**
I really don't think that my OS is the problem.

vilhelmprytz:
Fixes #112

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,76 +72,10 @@ Here is a [YouTube video](https://www.youtube.com/watch?v=E8UJhyUFoHM) that illu
 
 ## Firewall setup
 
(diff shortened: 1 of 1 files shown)


vilhelmprytz:
@Linux123123

PR #118, opened 2020-11-17
wings: Fix bugs, move functions around, use more case

Linux123123:
Fixed bugs, used a few more `case` instead of `if`. Added `User input functions`.
I had some code written already, before I saw the bugs, so I included it here too.


vilhelmprytz:
I'd like all submitted PRs to solve one thing (per PR). Are you able to just open a PR with the bug-fixes? Since it's a bit critical I want to merge it ASAP and I don't have time to review this at the moment :slightly_smiling_face:

PR #119, opened 2020-11-17
panel: fix broken if statement in perform_install function

Linux123123:
Fix 1 bug

### DIFF
--- a/install-wings.sh
+++ b/install-wings.sh
@@ -458,8 +458,8 @@ letsencrypt() {
 perform_install() {
   echo "* Installing pterodactyl wings.."
   [ "$OS" == "ubuntu" ] || [ "$OS" == "debian" ] && apt_update
-  [ "$OS" == "centos" ] || [ "$OS_VER_MAJOR" == "7" ] && yum_update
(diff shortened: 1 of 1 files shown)


install-wings.sh
```diff
@@ -458,8 +458,8 @@ letsencrypt() {
 perform_install() {
   echo "* Installing pterodactyl wings.."
   [ "$OS" == "ubuntu" ] || [ "$OS" == "debian" ] && apt_update
```
vilhelmprytz, line 460: Are you sure this works as expected? I don't think this is equivalent to

```bash
if [ "$OS" == "ubuntu" ] || [ "$OS" == "debian" ]; then
    apt_update
fi
```

but correct me if I am wrong.

Linux123123:
```bash
false || true && echo "Works"
true || false && echo "Works"
false || false && echo "Doesn't work"
true || true && echo "Works"
```
```bash
Works
Works
Works
```
Based on this it works. But I would also be skeptic about this if I wouldn't have tested it before.

vilhelmprytz (approved):

PR #123, opened 2020-12-10
Uninstall script

### LINKED ISSUE #141: A way to easily uninstall Pterodactyl and Wings
I sometimes need to uninstall pterodactyl but i do not know what files to uninstall because I use this installer. 

I think that a good solution for this is to have the option to uninstall all the pterodactyl and wings files using the pterodactyl installer.

I have considered wiping the machine, but that can get very tedious.

Linux123123:
Added an uninstall script. It should remove panel and wings without a need to reinstall OS. 

Fixes #141


vilhelmprytz:
I don't know if we want to encourage using an uninstallation script if it doesn't 100% cleanly revert to the initial state. At that point, why not use something like Ansible?

Linux123123:
I am not creating this to fully uninstall the panel /wings. I am creating this for people who want to reinstall pterodactyl after a failed installation, who don't want to reinstall OS.

uninstall.sh
```diff
@@ -0,0 +1,290 @@
...
+    echo "$valid_users"
+  fi
+  while [ -z "$USER" ] || [[ $valid_users != *"$user_input"* ]]; do
+    echo -n "* Choose the panel user(to skip type none): "
+    read -r user_input
+    USER=$user_input
+  done
+  [[ "$USER" != "none" ]] && mysql -u root -p -e "DROP USER '$USER'@'127.0.0.1'"
+  mysql -u root -p -e "FLUSH PRIVILEGES;"
+}
```
Linux123123, line 222: @vilhelmprytz Do you think database removal part is good enough to not confuse users?

vilhelmprytz: I am skepictal about removing all users but `root`. An ignorant user could easily accidentally wipe databases and users they do not wish to remove. Would it be possible to follow the same kind of "prompt format" as `install-panel.sh`? i.e. ask for MySQL username and database with the default values `panel` and `pterodactyl`.

Also, the uninstall script should be very clear about the fact that it is only meant to be run on systems that used the installation script in the first place. Haven't looked at your code yet so I don't know if you've made this apparent already. We could make the panel script create a file somewhere on the system when the install completes indicating that the system was installed using pterodactyl-installer and not with any other script or manually.

Linux123123: No it doesn't delete all users. It gets a list of users, excludes `root` because it can't be deleted. Then it searches for user pterodactyl in that list and if it is there it asks if it is the pterodactyl user and if answer is yes it deletes it. If answer is no it let's the user choose the user to be deleted from the list. Exactly the same is done with database.

I was thinking something better. A file with variables inside which indicate how much of the script has ran and based on that how much to remove. Most users run the script, then fails on something like database user name or pterodactyl admin account username and they need to wipe their system. This uninstall script is for those people.

vilhelmprytz:
What's the status of this PR?

Linux123123:
I am pretty sure I am waiting for your response if it is a good way to implement this.

vilhelmprytz:
I've read through the previous discussion we've had. Perhaps it's a bit too complicated to create a file on the server in order to indicate how much of the installation is completed. Just make sure the uninstall script can handle all errors, i.e. trying to remove stuff that's not there should not exit the script.

uninstall.sh
```diff
@@ -0,0 +1,291 @@
...
+  output
+  output "Copyright (C) 2018 - 2021, Vilhelm Prytz, <vilhelm@prytznet.se>"
+  output "https://github.com/vilhelmprytz/pterodactyl-installer"
+  output
+  output "Sponsoring/Donations: https://github.com/vilhelmprytz/pterodactyl-installer?sponsor=1"
+  output "This script is not associated with the official Pterodactyl Project."
+  output
+  output "Running $OS version $OS_VER."
+  print_brake 70
+  # check_os_comp
```
vilhelmprytz, line 248: Why is this commented?

uninstall.sh
```diff
@@ -0,0 +1,291 @@
...
+
+  if [ "$RM_PANEL" == false ] && [ "$RM_WINGS" == false ]; then
+    error "Nothing to uninstall!"
+    exit 1
+  fi
+
+  summary
+
+  # confirm uninstallation
+  echo -e -n "* Continue with installation? (y/N): "
```
vilhelmprytz, line 273:
```suggestion
  echo -e -n "* Continue with uninstallation? (y/N): "
```

uninstall.sh
```diff
@@ -0,0 +1,291 @@
...
+  read -r CONFIRM
+  if [[ "$CONFIRM" =~ [Yy] ]]; then
+    perform_uninstall
+  else
+    error "Installation aborted."
```
vilhelmprytz, line 278:
```suggestion
    error "Uninstallation aborted."
```

uninstall.sh
```diff
@@ -0,0 +1,291 @@
...
+    exit 1
+  fi
+}
+
+goodbye() {
+  print_brake 62
+  output "Panel uninstallation completed"
```
vilhelmprytz, line 285: So it will always print "panel uninstallation completed" even if only Wings was removed?

vilhelmprytz (commented):
Left some comments. I think it's good that you have decided not to uninstall any `apt` or `dnf` dependencies since the script would still succeed if it was run again on the same machine, right?

One thing though, if there are Wings servers running in Docker, what happens to those containers?

uninstall.sh
Linux123123: Testing purposes :) thanks for catching.

uninstall.sh
Linux123123: Made it so it changes based on user selection.

uninstall.sh
```diff
@@ -0,0 +1,292 @@
...
+  centos)
+    systemctl disable --now redis
+    systemctl disable --now php-fpm
+    rm -rf /etc/php-fpm.d/www-pterodactyl.conf
+    ;;
+  esac
+}
+
+rm_cron() {
+  # shellcheck disable=SC2063
```
vilhelmprytz, line 184: Why is this disabled with shellcheck?

uninstall.sh
```diff
@@ -0,0 +1,292 @@
...
+    echo "$valid_db"
+  fi
+  while [ -z "$DATABASE" ] || [[ $valid_db != *"$database_input"* ]]; do
+    echo -n "* Choose the panel database(to skip type none): "
+    read -r database_input
+    DATABASE=$database_input
+  done
+  [[ "$DATABASE" != "none" ]] && mysql -u root -p -e "DROP $DATABASE"
+  # Exclude usernames User and root (Hope no one uses username User)
+  valid_users=$(mysql -u root -p -e "SELECT user FROM mysql.user;" | grep -v -E -- 'User|root')
```
vilhelmprytz, line 205: Why are we excluding `User`? What's so specific about that user?

uninstall.sh
```diff
@@ -0,0 +1,292 @@
...
+  warning "Be careful! This user will be deleted!"
+  if [[ "$valid_users" == *"pteroactyl"* ]]; then
+    echo -n "* User called pterodactyl has been detected. Is it the pterodactyl user? (y/N): "
+    read -r is_user
+    [[ "$is_user" =~ [Yy] ]] && USER=pterodactyl || echo "$valid_users"
+  else
+    echo "$valid_users"
+  fi
+  while [ -z "$USER" ] || [[ $valid_users != *"$user_input"* ]]; do
+    echo -n "* Choose the panel user(to skip type none): "
```
vilhelmprytz, line 215: Perhaps it would be better to say to the user that `pterodactyl` MySQL user was not found, and then ask whether they would like to enter a specific username or if they want to skip this step (y/N prompt might be easier and less prone to cause user confusion, I first thought `none` meant typing nothing).

uninstall.sh
```diff
@@ -0,0 +1,292 @@
...
+    output "Panel installation has been detected."
+    echo -e -n "* Do you want to remove panel? (y/N):"
+    read -r RM_PANEL_INPUT
+    [[ "$RM_PANEL_INPUT" =~ [Yy] ]] && RM_PANEL=true
+  fi
+
+  if [ -d "/etc/pterodactyl" ]; then
+    output "Wings installation has been detected."
+    warning "This will remove all the servers!"
+    echo -e -n "* Do you want to remove wings(daemon)? (y/N):"
```
vilhelmprytz, line 260:
```suggestion
    echo -e -n "* Do you want to remove Wings (daemon)? (y/N):"
```

Should it be capitalized, what do you think?

vilhelmprytz (commented):
Great work, left some comments!

uninstall.sh
Linux123123: So you want that if the user `pterodactyl` is not found it would ask if user wants to select a user to delete and only then proceed to giving them selections?

uninstall.sh
Linux123123: When you use the `SELECT user FROM mysql.user;` it gives back a table with a column called user. So it needs to be excluded as it will be always when you run that command. But I just noticed, that I wrote it `User` not `user`.

uninstall.sh
Linux123123: This is because shellcheck when it sees a star thinks it is a glob. The `-F` flag fixes it , but I am not sure if it works with `-F`

uninstall.sh
Linux123123: Yeah sure, why not.

Truyn:
How do I use this script?
I do not understand

vilhelmprytz:
> How do I use this script?
> I do not understand

It is not yet merged, so not released yet.

UppishBuzzard87:
> > How do I use this script?
> > I do not understand
> 
> It is not yet merged, so not released yet.

Is this script actually safe to use yet, just awaiting a merge, or are there still bugs that makes this script unusable.

vilhelmprytz:
It probably works as intended, but I am yet to test it fully.

UppishBuzzard87:
Not sure if this is an error, but when I type `none` to skip the deletion of the database, it just asks the question again. I checked the code and it said ` [[ "$DATABASE" != "none" ]] && mysql -u root -p -e "DROP $DATABASE"` which drops the provided database if the answer wasn't 'none'. But instead for me, it just asks the question again.

image

Linux123123:
> Not sure if this is an error, but when I type `none` to skip the deletion of the database, it just asks the question again. I checked the code and it said ` [[ "$DATABASE" != "none" ]] && mysql -u root -p -e "DROP $DATABASE"` which drops the provided database if the answer wasn't 'none'. But instead for me, it just asks the question again.
> 
> image

This is being discussed [here](https://github.com/vilhelmprytz/pterodactyl-installer/pull/123#discussion_r640877979). I am intentionally not fixing it, because I am waiting for the answer. This is still WIP so you shouldn't have expected this to fully work.

UppishBuzzard87:
> > Not sure if this is an error, but when I type `none` to skip the deletion of the database, it just asks the question again. I checked the code and it said ` [[ "$DATABASE" != "none" ]] && mysql -u root -p -e "DROP $DATABASE"` which drops the provided database if the answer wasn't 'none'. But instead for me, it just asks the question again.
> > image
> 
> This is being discussed [here](https://github.com/vilhelmprytz/pterodactyl-installer/pull/123#discussion_r640877979). I am intentionally not fixing it, because I am waiting for the response. This is still WIP so you shouldn't have expected this to fully work.

Yeah, I didn't expect it to work. Thanks for the explanation. I will try this again when it's ready :)

uninstall.sh
vilhelmprytz: > So you want that if the user `pterodactyl` is not found it would ask if user wants to select a user to delete and only then proceed to giving them selections?

Yes.

Linux123123: It should work as intended now after https://github.com/vilhelmprytz/pterodactyl-installer/pull/123/commits/c252eb8817df079613bd5666e46860a62e059321

uninstall.sh
Linux123123: Just checked `-F` is sufficient enough.

Linux123123:
Ok now it should work as intended. Tried on `vagrant` (`ubuntu_focal`)  right after installing the panel, and it worked. Then after uninstalling tried installing the panel again and it also worked fine. :)

uninstall.sh
```diff
@@ -0,0 +1,326 @@
+#!/bin/bash
+
+set -e
```
vilhelmprytz, line 3: We discussed this on Discord. What are the pros and cons of using `set -e` for the uninstallation script? I can think of a few myself but I am open for dicussion.

uninstall.sh
```diff
@@ -0,0 +1,326 @@
...
+  output "Succesfully removed services."
+}
+
+rm_cron() {
+  output "Removing cron jobs..."
+  crontab -l | grep -vF "* * * * * php /var/www/pterodactyl/artisan schedule:run >> /dev/null 2>&1" | crontab -
+  output "Succesfully removed cron jobs."
+}
+
+rm_database() {
```
vilhelmprytz, line 204: I think we discussed this before as well, but do we want to uninstall MySQL/MariaDB completely? Since it can contain other data, I think we agreed that it is best to keep it installed. Do I remember correctly? Installing Pterodactyl again after using the uninstallation script should just work.

Linux123123: The script doesn't uninstall MariaDB, it only deletes the panel database and user. Yes completely uninstalling MariaDB would be really bad.

vilhelmprytz: Good.

uninstall.sh
```diff
@@ -0,0 +1,326 @@
...
+
+rm_wings_files() {
+  output "Removing wings files..."
+  rm -rf /etc/pterodactyl /usr/local/bin/wings /var/lib/pterodactyl
+  output "Succesfully removed wings files."
+}
+
+rm_services() {
+  output "Removing services..."
+  systemctl disable --now mariadb
```
vilhelmprytz, line 182: Should we really disable MariaDB? What if the user is using it for something else? I reckon we should just remove the database, but not stop the service itself.

vilhelmprytz (changes requested):
I have now tested this locally on Ubuntu and Debian. Code-wise, it looks very good. But I did notice some things when testing. I left a comment.

Also, the Wings service isn't removed. I tried starting Wings and I created a test server. Then I ran the uninstallation script. The server was still running after the script was completed. I am not sure how we will handle Docker containers, but the Wings service should be stopped and removed.

Anyways, apart from these two things, we should be good for merge (after testing on CentOS and re-testing on Ubuntu/Debian as well).

vilhelmprytz:
Also, another thing! Perhaps we should re-link `/etc/nginx/sites-available/default` to `/etc/nginx/sites-enabled/default` and restart nginx?

vilhelmprytz:
TODO: Add support for removing old Let's Encrypt certificates.

vilhelmprytz (approved):
This isn't fully ready yet, but I will merge it to get rid of the dangling PR. Thanks for the contribution as always @Linux123123!

PR #139, opened 2021-01-10
wings: use bracket statements

Linux123123:
Changed all 
```bash
"$var"
```
to 
```bash
[ "$var" == true ]
```


vilhelmprytz:
I've noticed that `[ "$var" == true ]` does not work on `zsh`. Do you know why?

Linux123123:
> I've noticed that `[ "$var" == true ]` does not work on `zsh`. Do you know why?

Not really.

PR #140, opened 2021-01-14
panel: remove ability to run p:environment:mail, redundant

### LINKED ISSUE #137: panel: question about email sending
**Describe the bug**
So basically we can skip the question about sending emails, because if we set `APP_ENVITONMENT_ONLY=false` we can have the user change it from panel. 

**To Reproduce**
1. Run the script
2. Choose panel
3. Get the question

**Expected behaviour**
It shouldn't ask for it.

**Screenshots**
image

 - OS: All
 - Installation: panel

**Additional context**
It is a lot easier to set them from within the panel. Also most users set the `APP_ENVITONMENT_ONLY=false` anyway, so the script can automatically do it.

vilhelmprytz:
Fixes #137

### DIFF
--- a/install-panel.sh
+++ b/install-panel.sh
@@ -396,9 +396,6 @@ configure() {
     --username="$MYSQL_USER" \
     --password="$MYSQL_PASSWORD"
 
-  # Email credentials manually set by user
-  [[ "$mailneeded" =~ [Yy] ]] && php artisan p:environment:mail
-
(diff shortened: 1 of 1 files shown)


vilhelmprytz:
@Linux123123 Something like this?

Linux123123:
> @Linux123123 Something like this?

Yes.

PR #150, opened 2021-01-27
wings: centos: replace pterodactyl0 with docker0 trusted zone

vilhelmprytz:
I've had some issues where Wings won't boot on CentOS 7/8 because the interface "isn't trusted" or something like that. Only occurs sometimes for me, but I thought it's worth looking into.


vilhelmprytz:
I noticed I forgot the `-q`. Will add back.

vilhelmprytz:
This is broken and most likely not something people are having problems with. I'll just create a new PR if I need to change it.

PR #166, opened 2021-02-24
panel: User based on OS pteroq

### LINKED ISSUE #165: Pteroq service has wrong users in CentOS7
See this bug https://github.com/pterodactyl/panel/issues/2806

When I used this script it set the user as `www-data` in the `pteroq.serivce` file causing the service to fail. CentOS7 uses `nginx` as the user.

Linux123123:
Based on OS changes the user in `pteroq` service file.

Fixes #165


vilhelmprytz (approved):
Thanks for contributing, LGTM!

PR #176, opened 2021-04-13
Fix wings configuring link

sinmineryt:
I changed the link that appears when the installation ends. 
When clicking on the old link (https://pterodactyl.io/wings/1.0/installing.html#configure-daemon) It would put you to the start of the page. 
When clicking the new link (https://pterodactyl.io/wings/1.0/installing.html#configure), it would directly put you to the part where you configure the daemon. 
It could confuse some users that they should reinstall wings when clicking the old link.


vilhelmprytz (approved):
LGTM

PR #181, opened 2021-05-16
panel/wings: Add email validation

### LINKED ISSUE #157: panel: check email address provided
When you enter an email address, it isn't checked and just passed to `letsencrypt` and `artisan`. It can be invalid and cause problems.

Check the input with this regex:
```
/^[^\s@]+@[^\s@]+\.[^\s@]+$/
```

Entering email address that is correct. :)

Linux123123:
Added email validation with regex. I think it works great, as far as I tested.

Fixes #157

### DIFF
--- a/install-panel.sh
+++ b/install-panel.sh
@@ -108,7 +108,7 @@ array_contains_element() {
   return 1
 }
 
-valid_email () {
+valid_email() {
   [[ $1 =~ ${regex} ]]
 }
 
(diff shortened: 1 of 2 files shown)


vilhelmprytz (commented):
I think this looks good, but does it handle all valid emails? See [this article](https://en.wikipedia.org/wiki/Email_address#Local-part), I think some characters are classified as invalid by this regex. Worth bothering or not?

Linux123123:
No, it's not worth bothering as this regex is compliant with most of the email standards. Either way we only want to check more of the structure of them email, then the email characters. I think just leave this regex as is and it should work.

vilhelmprytz (approved):
LGTM

PR #188, opened 2021-05-21
wings: Automatically skip MariaDB question if MySQL/MariaDB is detected

### LINKED ISSUE #81: Automatically skipping question if panel/database has been installed while installing Wings
Some people just don't read text above a question clearly.

Automatically skip the question if a MySQL instance is detected on the system.

None.

**Additional context**
Small addition to the script. After you have installed only the panel and want to install the Wings using the same script. It tells you that if you have the panel installed you should answer no to the next question. But can't this just be skipped since you can just check if a panel has been installed and then skip that option for the user by default?

Referring to when it asks if you want to install MariaDB in the Wings script.

vilhelmprytz:
Fixes #81

### DIFF
--- a/install-wings.sh
+++ b/install-wings.sh
@@ -530,11 +530,14 @@ main() {
   echo -e "* ${COLOR_RED}Note${COLOR_NC}: this script will not enable swap (for docker)."
   print_brake 42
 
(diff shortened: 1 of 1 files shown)


vilhelmprytz:
Tested to work.

PR #202, opened 2021-07-01
Allow CNAME DNS records

### LINKED ISSUE #201: CNAME records not supported
Using CNAME records results in the script claiming a domain does not point to the proper IP address.

Support CNAME records.

None.

**Additional context**
Ran the script on a clean Debian 10 installation

This is the logging for an example where sub.primarydomain.com points to sub2.diffrentdomain.com, which points to sub3.duckdns.org, which points to 10.0.0.22:

```
* I agree that this HTTPS request is performed (y/N): y
* Resolving DNS for sub.primarydomain.com
* The DNS record (sub2.diffrentdomain.com.
sub3.duckdns.org.
10.0.0.22) does not match your server IP. Please make sure the FQDN games.wehlan.nl is pointing to the IP of your server, 10.0.0.22
* If you are using Cloudflare, please disable the proxy or opt out from Let's Ecnrypt.
* Proceed anyways (your install will be broken if you do not know what you are doing)? (y/N):
```

Please note that the IP does actually show up in the DNS record, alongside the CNAME records.

jobhh:
Disclaimer: I have only tested this in an online bash shell, please check if it properly works.

This should fix the situation where a CNAME record is used to point to a domain that points to the proper IP address.
It does break when someone uses a domain that contains the IP address but does not point to the IP address, but I think that's an bigger edge case than using CNAME records.

I'm open for other suggestions 😃 

Fixes #201

### DIFF
--- a/lib/verify-fqdn.sh
+++ b/lib/verify-fqdn.sh
@@ -90,7 +90,7 @@ dns_verify() {
   output "Resolving DNS for $fqdn"
   ip=$(curl -4 -s $CHECKIP_URL)
   dns_record=$(dig +short @$DNS_SERVER "$fqdn")
-  [[ $dns_record != *"$ip"* ]] && fail
+  [ "${ip}" != "${dns_record} | tail -n1" ] && fail
   output "DNS verified!"
 }
 
--- a/lib/verify-fqdn.sh
+++ b/lib/verify-fqdn.sh
@@ -90,7 +90,7 @@ dns_verify() {
   output "Resolving DNS for $fqdn"
   ip=$(curl -4 -s $CHECKIP_URL)
   dns_record=$(dig +short @$DNS_SERVER "$fqdn")
-  [ "${ip}" != "${dns_record}" ] && fail
+  [[ $dns_record != *"$ip"* ]] && fail
   output "DNS verified!"
 }
 
--- a/lib/verify-fqdn.sh
+++ b/lib/verify-fqdn.sh
@@ -89,8 +89,8 @@ confirm() {
 dns_verify() {
   output "Resolving DNS for $fqdn"
   ip=$(curl -4 -s $CHECKIP_URL)
-  dns_record=$(dig +short @$DNS_SERVER "$fqdn")
-  [ "${ip}" != "${dns_record} | tail -n1" ] && fail
+  dns_record=$(dig +short @$DNS_SERVER "$fqdn" | tail -n1)
+  [ "${ip}" != "${dns_record}" ] && fail
   output "DNS verified!"
 }
 


lib/verify-fqdn.sh
```diff
@@ -90,7 +90,7 @@ dns_verify() {
   output "Resolving DNS for $fqdn"
   ip=$(curl -4 -s $CHECKIP_URL)
   dns_record=$(dig +short @$DNS_SERVER "$fqdn")
-  [ "${ip}" != "${dns_record}" ] && fail
+  [[ $dns_record != *"$ip"* ]] && fail
```
vilhelmprytz, line 93: Why have they switched places?

vilhelmprytz (commented):
Thank you for contributing! As far as I can see, `dig +short @8.8.8.8 "$fqdn"` where `fqdn` is a CNAME will print "two lines". First, the value of the CNAME record. Secondly, the A record (IPv4) of the record that the CNAME points to. I think it would be better to check if the string `dns_record` has two lines and then check if the second line matches the IP we are looking for (and if it does not have two lines, then simply perform the current check). Do you agree that this would handle all use cases and do you think it is something you could implement? I am open to discussion about the implementation 🙂

lib/verify-fqdn.sh
jobhh: No real reason, I just quickly typed that. 
I hardly ever type shell, but I think the "'s can also be removed around $ip.

jobhh:
> 
> 
> Thank you for contributing! As far as I can see, `dig +short @8.8.8.8 "$fqdn"` where `fqdn` is a CNAME will print "two lines". First, the value of the CNAME record. Secondly, the A record (IPv4) of the record that the CNAME points to. I think it would be better to check if the string `dns_record` has two lines and then check if the second line matches the IP we are looking for (and if it does not have two lines, then simply perform the current check). Do you agree that this would handle all use cases and do you think it is something you could implement? I am open to discussion about the implementation 🙂

I think it's safe to assume the last line of the dns response is always the ip address, but we cannot assume it will be 1 or 2 lines. If a CNAME record is pointing towards another CNAME record, it will already be 3 lines.

In my case it printed 3 lines. I have a domain specific to the pterodactyl service pointing to the personalised domain of my home address, which points to the domain of my dynamic dns provider (DuckDNS), which points to the ip address of my home. This extra CNAME record allows me to only change 1 DNS record if I ever have to switch away from DuckDNS.

In my example this is the reply from the DNS check for `sub.primarydomain.com`:
```
sub2.diffrentdomain.com.
sub3.duckdns.org.
10.0.0.22
```

I'll make a check for the last line.

vilhelmprytz (approved):
LGTM. I moved the `tail` to the actual variable definition. I've tried this with normal A records, CNAME records pointing to A records, and CNAME records pointing to another CNAME record that then points to an A record - it all works as expected. Thank you for suggesting and implementing this!

PR #203, opened 2021-07-03
Support installing Wings on arm64

### LINKED ISSUE #204: wings: support arm64
Since Wings is released as an arm64 version and Docker supports it (although not all Docker images), it could be implemented.

Self-explanatory.

N/A

**Additional context**
Thank you @puiemonta1234 for suggesting and implementing!

puiemonta1234:
Fixes #204

### DIFF
--- a/install-wings.sh
+++ b/install-wings.sh
@@ -172,8 +172,8 @@ check_os_comp() {
   x86_64)
     ARCH=amd64
     ;;
-  arm64)
-    ;& # fallthrough
+  arm64) ;&
+    # fallthrough
   aarch64)
     print_warning "Detected architecture arm64"
     print_warning "You will need to use Docker images made specifically for arm64"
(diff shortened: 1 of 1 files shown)


vilhelmprytz (changes requested):
Thank you for contributing! I definitely think this is something we should implement. MariaDB does not support arm64, the Wings script needs to skip the option to install it when the user uses arm64. Are you able to implement this?

puiemonta1234:
Actually, I had no problems using install-panel.sh on arm64, which installs MariaDB from the official repos (Ubuntu 20.04 in my case). Also, the script that install-wings.sh uses (the one from downloads.mariadb.com) states that "Only Ubuntu 16/xenial, 18/bionic, and 20/focal are supported for ARM64" and "Only RHEL/CentOS 7 are supported for ARM64", so it isn't always the case that MariaDB doesn't support arm64. We could detect the error from MariaDB's install script and abort, or install it from the official repos, like install-panel.sh does.

vilhelmprytz (approved):
LGTM!

PR #222, opened 2021-08-01
Database hosts

### LINKED ISSUE #220: Suggestion: Automatically create user for database host
X

When asking for the database username, also ask if the user would want to set up another user for pterodactyl database hosts, if yes, ask for the username.

X

**Additional context**
This will make it easier for the user to create a database host.

sinjs:
This pull request adds the automatic configuration of database hosts to the script.
It will ask you for the username and password of the database user and then create the user.

Fixes #220

### DIFF
--- a/install-wings.sh
+++ b/install-wings.sh
@@ -72,6 +72,11 @@ CONFIGURE_LETSENCRYPT=false
 FQDN=""
 EMAIL=""
 
+# Database host
+CONFIGURE_DBHOST=false
+MYSQL_DBHOST_USER="pterodactyluser"
+MYSQL_DBHOST_PASSWORD="password"
+
 # regex for email input
 regex="^(([A-Za-z0-9]+((\.|\-|\_|\+)?[A-Za-z0-9]?)*[A-Za-z0-9]+)|[A-Za-z0-9]+)@(([A-Za-z0-9]+)+((\.|\-|\_)?([A-Za-z0-9]+)+)*)+\.([A-Za-z]{2,})+$"
 
@@ -92,6 +97,42 @@ WINGS_VERSION="$(get_latest_release "pterodactyl/wings")"
 
 valid_email() {
   [[ $1 =~ ${regex} ]]
+}
+
+password_input() {
+  local __resultvar=$1
+  local result=''
+  local default="$4"
+
+  while [ -z "$result" ]; do
+    echo -n "* ${2}"
+
+    # modified from https://stackoverflow.com/a/22940001
+    while IFS= read -r -s -n1 char; do
+      [[ -z $char ]] && {
+        printf '\n'
+        break
+      }                               # ENTER pressed; output \n and break.
+      if [[ $char == $'\x7f' ]]; then # backspace was pressed
+        # Only if variable is not empty
+        if [ -n "$result" ]; then
+          # Remove last char from output variable.
+          [[ -n $result ]] && result=${result%?}
+          # Erase '*' to the left.
+          printf '\b \b'
+        fi
+      else
+        # Add typed char to output variable.
+        result+=$char
+        # Print '*' in its stead.
+        printf '*'
+      fi
+    done
+    [ -z "$result" ] && [ -n "$default" ] && result="$default"
+    [ -z "$result" ] && print_error "${3}"
+  done
+
+  eval "$__resultvar="'$result'""
 }
 
 #################################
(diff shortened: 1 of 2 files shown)


Linux123123:
Shouldn't this be in `install-wings.sh`? There is not a whole lot of reasons to have database host without `wings`.

sinmineryt:
> Shouldn't this be in `install-wings.sh`? There is not a whole lot of reasons to have database host without `wings`.

Yeah, correct. I'll move the code.

sinmineryt:
And I moved the code

sinmineryt:
Oh wait I forgot something

Linux123123:
I think it should also automatically add the necessary config for `mariaDB` to bind to `0.0.0.0` (and probably warn about that).

sinmineryt:
> I think it should also automatically add the necessary config for `mariaDB` to bind to `0.0.0.0` (and probably warn about that).

How would I do that? The line is not necessarily present in the file.

sinmineryt:
I'm currently trying to run the same command on CentOS 8, but my computer doesn't want to :^(

sinmineryt:
On CentOS 8, there isn't even an `/etc/mysql` folder.

sinmineryt:
For CentOS 8, this command should work: `sed -e 's/^#bind-address=0.0.0.0$/bind-address=0.0.0.0/' /etc/my.cnf.d/mariadb-server.cnf`

install-wings.sh
```diff
@@ -558,6 +634,19 @@ main() {
   $ASK_MYSQL && read -r CONFIRM_INSTALL_MARIADB
   $ASK_MYSQL && [[ "$CONFIRM_INSTALL_MARIADB" =~ [Yy] ]] && INSTALL_MARIADB=true
 
+  if [ "$INSTALL_MARIADB" == true ]; then
+    ask_database_host
+  fi
+
+  if [ "$CONFIGURE_DBHOST" == true ]; then
+    echo -n "* Database host username (pterodactyluser): "
```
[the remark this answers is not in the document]
sinmineryt: No it wouldn't. I'll implement this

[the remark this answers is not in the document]
sinmineryt: Actually, It will not work. Since this will only run if it did not already detect MySQL, and if there is no MySQL, there cannot be any user.

install-wings.sh
```diff
@@ -558,6 +634,19 @@ main() {
...
+    read -r MYSQL_DBHOST_USER_INPUT
+    
+    [ -z "$MYSQL_DBHOST_USER_INPUT" ] && MYSQL_DBHOST_USER="pterodactyluser" || MYSQL_DBHOST_USER=$MYSQL_DBHOST_USER_INPUT
```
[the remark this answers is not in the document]
sinmineryt: Correct.

install-wings.sh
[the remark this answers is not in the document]
sinmineryt: But I've added the commit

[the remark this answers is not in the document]
Linux123123: I think that you should also ask about database hosts when `ASK_MYSQL` is false. Then there would be a reason to check for existing users.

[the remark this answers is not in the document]
sinmineryt: Okay I will do that.

install-wings.sh
```diff
@@ -558,6 +634,29 @@ main() {
   $ASK_MYSQL && read -r CONFIRM_INSTALL_MARIADB
   $ASK_MYSQL && [[ "$CONFIRM_INSTALL_MARIADB" =~ [Yy] ]] && INSTALL_MARIADB=true
```
vilhelmprytz, line 635: Is it necessary to first ask if the user wants MariaDB installed, and then asking if they want the automatic setup? It seems useless to want the script to install MariaDB, but not do anything about it. What do you reckon?

vilhelmprytz (commented):
One comment! Will also test this locally ASAP.

install-wings.sh
sinjs: I think it is necessary because why would you want to automatically configure database hosts if you don't want MariaDB installed?

> It seems useless to want the script to install MariaDB, but not do anything about it.

Yeah, it is useless. But I'm still letting the user decide if they want it to be configured or not.

vilhelmprytz: Sorry, I realised I might have sounded a bit confused. I suggest just simply asking: do you want to automatically configure a database user? If the server has MariaDB, then just create the user. If the server does not have MariaDB, then install it and create the user. 🙂

vilhelmprytz: Do you reckon that’s a good solution?

sinjs: I like both solutions, mine and yours.

vilhelmprytz: Would there be a reason for someone to want MariaDB just installed and nothing else?

sinjs: Well, actually, if I think about it: No. I'll implement your solution

sinjs: Implemented.

sinjs: And I forgot to sign my commit again :^)

install-wings.sh
```diff
@@ -504,6 +579,7 @@ perform_install() {
   ptdl_dl
   systemd_file
   [ "$INSTALL_MARIADB" == true ] && install_mariadb
```
vilhelmprytz, line 581: Do you know if `install_mariadb` handles to be run twice (e.g. once for panel installation and once for Wings installation, if running both on the same machine)?

vilhelmprytz (approved):
Thank you for contributing. I have tested this locally now, works as expected. LGTM! :+1:

PR #228, opened 2021-08-13
[phpmyadmin install script]

Ferks-FK:
### **Attention!**
First of all, this fork was made for personal use purposes, so there are a lot of modifications.
I'm just making this request to share the phpmyadmin install script Intending to help.

_If possible, only take into account the phpmyadmin installation scripts, the other changes are not important._

Here is a small phpmyadmin install script for the pterodactyl panel.

### **Final considerations:**
I couldn't find a very user-friendly way to solve the problem in the image below, this I leave with someone more knowledgeable in shell.
Sem título

### **EDIT:**
The above image issues have already been fixed after good code changes.


Linux123123:
Thanks for the PR. :)

So first of all I see that you are just downloading the `phpmyadmin` zip and unzipping it straight into the `public` directory. I think the better way to do it is to use the package provided by the OS. Then (in debian based OS's) it can easily be configured when installing the package with `dbconf`. After that you would just add a location attribute to the `pterodactyl.conf`.

This way:
A - the package is properly configured
B - it can easily be updated

And cons:
CentOS doesn't have auto configuration, so wouldn't be supported / would need to be manually configured.

Well that is just my opinion. I will have to wait what @vilhelmprytz says

Ferks-FK:
> Obrigado pelo PR. :)
> 
> Em primeiro lugar, vejo que você está apenas baixando o `phpmyadmin`zip e descompactando-o diretamente no `public`diretório. Acho que a melhor maneira de fazer isso é usando o pacote fornecido pelo sistema operacional. Então (em sistemas operacionais baseados em debian) ele pode ser facilmente configurado ao instalar o pacote com `dbconf`. Depois disso, você apenas adicionaria um atributo de localização ao `pterodactyl.conf`.
> 
> Desta forma:
> A - o pacote está configurado corretamente
> B - pode ser facilmente atualizado
> 
> E contras: o
> CentOS não tem configuração automática, então não seria compatível / precisaria ser configurado manualmente.
> 
> Bem, essa é apenas minha opinião. Terei que esperar o que diz @vilhelmprytz


Um, true, it seems to be more interesting to do it the way you say.
Would it be something like apt-get install phpmyadmin?

Linux123123:
> Um, true, it seems to be more interesting to do it the way you say.
> Would it be something like apt-get install phpmyadmin?

Yes, but more complicated. I am working on it.

Ferks-FK (commented):
Fixed issue with phpmyadmin folders and permissions.

vilhelmprytz:
Thank you for submitting this PR. There are a lot of changes here, some of which I don't think are related to phpMyAdmin support? Would you be able to submit a PR with only the changes related to that feature implementation?

PR #231, opened 2021-08-15
Add debian 11 support

### LINKED ISSUE #230: Debian 11 Support
Debian 11 aka Bullseye was released yesterday.

Support for it would be great 👍🏻

Linux123123:
Adds Debian 11 support

Fixes #230

I think I changed everything, but if I missed something please tell.

Tested on Debian 11 vagrant (didn't test with let's encrypt)


vilhelmprytz (approved):
As always, thanks for the contribution! Tested locally, works. LGTM :+1:

PR #259, opened 2021-10-30
Update PROXMOX NESTING

Clashplayer-PROTECT:
FR : Juste la désactivation de la détection LCX car sur PROXMOX, on peut utiliser docker sur lcx avec l'option nesting in container.
EN : Just the deactivation of the LCX detection because on PROXMOX, we can use docker on lcx with nesting in container option.

Fixes #??

### DIFF
--- a/install-wings.sh
+++ b/install-wings.sh
@@ -293,7 +293,7 @@ check_os_comp() {
   virt_serv=$(virt-what)
 
   case "$virt_serv" in
-  *openvz* | *lxc*)
+  *openvz* )
     print_warning "Unsupported type of virtualization detected. Please consult with your hosting provider whether your server can run Docker or not. Proceed at your own risk."
     echo -e -n "* Are you sure you want to proceed? (y/N): "
     read -r CONFIRM_PROCEED


Linux123123:
Yes, but nesting is usually disabled in most `lxc` hostings. Just removing it from checks, doesn't make enabled all of a sudden. A better solution would be detecting if nesting is enabled in `lxc` and then bypassing this check.

Also you should open an issue about this.

vilhelmprytz:
As @Linux123123 has mentioned, we shouldn't remove this check. The script still lets you bypass the check if you know what you are doing. The only other solution I can think of is specifically detecting possibly supported scenarios where LXC and Docker will work, but I think that's a little bit unnecessary. Did you have problems with bypassing the check? @Clashplayer-PROTECT

Clashplayer-PROTECT:
I put my own patch on my github when I install on proxmox

vilhelmprytz:
Why not let the script run and "bypass" the check?

Clashplayer-PROTECT:
it's just for me personally

vilhelmprytz:
But why do you need your own patch when the script is capable of overriding the check out of the box?

Clashplayer-PROTECT:
It's just a person story

vilhelmprytz:
> It's just a person story

A person story?

PR #264, opened 2021-11-16
Fix incorrectly labeled setup option

NoahvdAa:
The setup option refers options five and six (legacy panel+wings and canary panel), this should be six and seven (canary panel and canary wings)


vilhelmprytz (approved):
Nicely found! Thanks for contributing 👍

PR #277, opened 2021-12-18
Fix database hosts

### LINKED ISSUE #267: Creating a Database Host Function
# [ Explanation ]
I would like to suggest an edit to the wings installation script.

Recently it was implemented the option for you to create the user for the database host creation automatically.

### [ How it currently works ]
An option appears to choose whether or not you want a database host user.
If they choose "Yes", they will be prompted for a user and password, and that is it.

### [ How it could work ]
Before the "User" and "Password" questions appear, ask the user whether the host will be functional only locally, or externally.

### [ Examples: ]

Would you like to create your database host automatically? "Yes"
Will your host be accessible Locally, or externally? ( Local/External )

If the answer is "Local", do it the original way.
If "External", ask for the host IP.

### [ Why this change? ]

Well, the script will not be limited to always creating hosts locally, and it will have an option for this to work in both cases.

Linux123123:
Fixes #267 

Also adds restarting of `mysqld` to apply config changes.

### DIFF
--- a/install-wings.sh
+++ b/install-wings.sh
@@ -485,7 +485,8 @@ ask_database_external() {
 }
 
 ask_database_firewall() {
-  echo -n "* Do you want to allow port 3306 threw the firewall? (y/N): "
+  print_warning "Allow incoming traffic to port 3306 (MySQL) can potentially be a security risk, unless you know what you are doing!"
+  echo -n "* Would you like to allow incoming traffic to port 3306? (y/N): "
   read -r CONFIRM_DB_FIREWALL
   if [[ "$CONFIRM_DB_FIREWALL" =~ [Yy] ]]; then
     CONFIGURE_DB_FIREWALL=true
--- a/install-wings.sh
+++ b/install-wings.sh
@@ -74,6 +74,8 @@ EMAIL=""
 
 # Database host
 CONFIGURE_DBHOST=false
+CONFIGURE_DBEXTERNAL=false
+CONFIGURE_DBEXTERNAL_HOST="%"
 MYSQL_DBHOST_USER="pterodactyluser"
 MYSQL_DBHOST_PASSWORD="password"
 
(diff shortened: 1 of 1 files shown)


Linux123123:
We may as well open the port if user selected to configure firewall and mysql to be external.

install-wings.sh
```diff
@@ -461,31 +463,59 @@ ask_database_user() {
...
-    ;;
-  centos)
-    sed -ne 's/^#bind-address=0.0.0.0$/bind-address=0.0.0.0/' /etc/my.cnf.d/mariadb-server.cnf
-    ;;
-  esac
+
+  if [ "$CONFIGURE_DBEXTERNAL" == true ]; then
+    case "$OS" in
+    debian | ubuntu)
+      sed -ne 's/^bind-address            = 127.0.0.1$/bind-address=0.0.0.0/' /etc/mysql/mariadb.conf.d/50-server.cnf
```
vilhelmprytz, line 510: I just tried this locally, and this doesn't actually change anything. Perhaps this would work as well (I have tried it locally on Debian 11)?
```suggestion
      sed -i 's/127.0.0.1/0.0.0.0/g' /etc/mysql/mariadb.conf.d/50-server.cnf
```

vilhelmprytz (changes requested):

install-wings.sh
Linux123123: Yes, this is a problem. It was added in #220 and I have also encountered it not working. Your solutions works a lot better.

Linux123123: Is this also a problem for CentOS?

vilhelmprytz (approved):
LGTM 👍. Would you like to add so that the script opens 3306 if the user opts in?

Linux123123:
> LGTM +1. Would you like to add so that the script opens 3306 if the user opts in?

Yes just added.

install-wings.sh
```diff
@@ -461,31 +464,68 @@ ask_database_user() {
...
+    if [ "$CONFIRM_DBEXTERNAL_HOST" != "" ]; then
+      CONFIGURE_DBEXTERNAL_HOST="$CONFIRM_DBEXTERNAL_HOST"
+    fi
+    [ "$CONFIGURE_FIREWALL" == true ] && ask_database_firewall
+    CONFIGURE_DBEXTERNAL=true
+  fi
+}
+
+ask_database_firewall() {
+  echo -n "* Do you want to allow port 3306 threw the firewall? (y/N): "
```
vilhelmprytz, line 488:
```suggestion
  print_warning "Allow incoming traffic to port 3306 (MySQL) can potentially be a security risk, unless you know what you are doing!"
  echo -n "* Would you like to allow incoming traffic to port 3306? (y/N): "
```

vilhelmprytz (approved):

vilhelmprytz:
Is it possible this PR solved #292 as well?

vilhelmprytz:
Nevermind my previous message, the issue I mentioned should not be related 🙂

PR #280, opened 2021-12-20
Updating the old PR.

Ferks-FK:
It is not ready yet, many things need to be done and tested.


Ferks-FK:
Regarding the latest update that was released, I am having a visual problem that I need help with.
The script checks that the user does not enter a password, so it returns an error, but even when you enter a password, it is called again.
Example:



Note: I know you need to do the password hiding function, but when I implement that, after you enter a password, the script doesn't respond, I believe it must loop infinitely on something.

vilhelmprytz:
Thank you for submitting this PR! The builds are failing, would you mind adjusting your PR so that the builds do not fail? Thanks! :slightly_smiling_face:

Ferks-FK:
> Thank you for submitting this PR! The builds are failing, would you mind adjusting your PR so that the builds do not fail? Thanks! 🙂

I honestly don't understand how this compilation works, can you explain it to me, or give me an example on how to fix it?

Linux123123:
It is not compilation. It is checking for common errors which may lead to bigger problems down the line.
When you click on the information of the build you will see
image
Here the errors are described. All of your errors appear to be related to not used variables. So to fix these problems you probably need to use the variable or delete it's definition.

Ferks-FK:
It requires some testing.

vilhelmprytz (changes requested):
Thank you for submitting this PR! 🙂 

I understand that you have built this as a separate "script", a new option in `install.sh`. This means that you have to reinvent a lot of the functions and processes already occurring in `install-panel.sh`. Would it not be better if support for installing phpMyAdmin was added as an option inside of the already existing installation script(s)?

Ferks-FK:
@vilhelmprytz Take a look at the new modification 😁

configs/pma.conf
```diff
@@ -0,0 +1,44 @@
+server {
```
Linux123123, line 1: Are you sure 2 more `nginx` configs are needed? Can't you just use the pterodactyl one and replace the root / other changes?

Ferks-FK: I believe it is good to leave this separate, after all, it is being installed in `/var/www/phpmyadmin`, not that it is mandatory, but it is good to separate.

Ferks-FK: Oops, I understood the question now 🤣
I'll check it out

vilhelmprytz (commented):
Shouldn't `install-pma.sh` be removed now since it looks like you've incorporated the installation of phpMyadmin inside of `install-panel.sh`?

Ferks-FK:
> Shouldn't `install-pma.sh` be removed now since it looks like you've incorporated the installation of phpMyadmin inside of `install-panel.sh`?

Yes, I left it there just for a kind of "backup".
Can I remove it?

vilhelmprytz:
> > Shouldn't `install-pma.sh` be removed now since it looks like you've incorporated the installation of phpMyadmin inside of `install-panel.sh`?
> 
> 
> 
> Yes, I left it there just for a kind of "backup".
> 
> Can I remove it?
> 
> 

Yes, you may remove it. The contents will still be available if looking at older commits, so it won't be completely lost.

Ferks-FK:
> > > Shouldn't `install-pma.sh` be removed now since it looks like you've incorporated the installation of phpMyadmin inside of `install-panel.sh`?
> > 
> > 
> > Yes, I left it there just for a kind of "backup".
> > Can I remove it?
> 
> Yes, you may remove it. The contents will still be available if looking at older commits, so it won't be completely lost.

Done 🤠

install-panel.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
+#shellcheck source=/dev/null
```
Ferks-FK, line 2: See: https://github.com/koalaman/shellcheck/wiki/SC1091

install-panel.sh
```diff
@@ -1176,6 +1146,34 @@ main() {
...
 
+  # Login and FQDN for phpmyadmin #
+  if [ "$INSTALL_PHPMYADMIN" == true ]; then
+    while [ -z "$PMA_FQDN" ] || [ "$FQDN" == "$PMA_FQDN" ]; do
+      echo -n "* Enter the FQDN for phpmyadmin (pma.example.com): "
+      read -r PMA_FQDN
+      [ -z "$PMA_FQDN" ] && print_error "FQDN cannot be empty"
+      [ "$FQDN" == "$PMA_FQDN" ] && print_error "The pterodactyl panels domain cannot be the same as PhpMyAdmins!"
+    done
+      if [ "$CONFIGURE_LETSENCRYPT" == true ]; then
```
Ferks-FK, line 1157: Added question to generate SSL certificates for the PhpMyAdmin domain

install-panel.sh
```diff
@@ -1209,6 +1207,9 @@ summary() {
   echo "* Configure Let's Encrypt? $CONFIGURE_LETSENCRYPT"
   echo "* Assume SSL? $ASSUME_SSL"
   echo "* Configure PhpMyAdmin? $INSTALL_PHPMYADMIN"
+  [ "$INSTALL_PHPMYADMIN" == true ] && echo "* Hostname/FQDN PhpMyAdmin: $PMA_FQDN"
```
Ferks-FK, line 1210: Added PMA summary, if true.

install-panel.sh
Linux123123: Not needed, as already ignored at the project level. Check `travis.yml`.

Ferks-FK: Ok.

install-panel.sh
```diff
@@ -804,6 +834,119 @@ letsencrypt() {
   fi
 }
 
+configure_pma() {
```
Linux123123, line 837: I think you should also split this function into multiple smaller functions and use the correct indentation of `2` spaces.

install.sh
```diff
@@ -82,6 +82,8 @@ PANEL_LATEST="$GITHUB_BASE_URL/$SCRIPT_VERSION/install-panel.sh"
 
 WINGS_LATEST="$GITHUB_BASE_URL/$SCRIPT_VERSION/install-wings.sh"
 
+PMA_LATEST="$GITHUB_BASE_URL/$SCRIPT_VERSION/install-pma.sh"
```
Linux123123, line 85: Also you are still making changes to `install.sh`, when they aren't needed.

install-panel.sh
Ferks-FK: Done

install-panel.sh
```diff
@@ -1030,6 +1209,10 @@ summary() {
   echo "* Configure Firewall? $CONFIGURE_FIREWALL"
   echo "* Configure Let's Encrypt? $CONFIGURE_LETSENCRYPT"
   echo "* Assume SSL? $ASSUME_SSL"
+  echo "* Configure PhpMyAdmin? $INSTALL_PHPMYADMIN"
+  [ "$INSTALL_PHPMYADMIN" == true ] && echo "* Hostname/FQDN PhpMyAdmin: $PMA_FQDN"
+  [ "$INSTALL_PHPMYADMIN" == true ] && echo "* PhpMyAdmin Username: $PMA_MYSQL_USER"
+  [ "$INSTALL_PHPMYADMIN" == true ] && echo "* phpmyadmin Password: $PMA_MYSQL_PASSWORD"
```
Linux123123, line 1215: Do not actually show the password. It's a security risk.

install.sh
```diff
@@ -109,6 +109,7 @@ while [ "$done" == false ]; do
     "$PANEL_LATEST"
     "$WINGS_LATEST"
     "$PANEL_LATEST;$WINGS_LATEST"
+    "$PMA_LATEST"
```
Linux123123, line 112: You left one line in `install.sh`

install-panel.sh
```diff
@@ -49,12 +49,23 @@ fi
 GITHUB_SOURCE="master"
 SCRIPT_RELEASE="canary"
 
+# Main Variables #
+
+#SUPPORT_LINK="https://pterodactyl-installer.se/discord"
```
Linux123123, line 54: What is the SUPPORT_LINK for? (even though it's commented).

install-panel.sh
```diff
@@ -117,6 +132,7 @@ valid_email() {
 print_error() {
   COLOR_RED='\033[0;31m'
   COLOR_NC='\033[0m'
+  COLOR_DEFAULT='\e[0m'
```
Linux123123, line 135: What is this used for?

install-panel.sh
```diff
@@ -787,8 +813,12 @@ letsencrypt() {
   # Obtain certificate
   certbot --nginx --redirect --no-eff-email --email "$email" -d "$FQDN" || FAILED=true
 
+  if [ "$CONFIGURE_SSL_PMA" == true ]; then
+    certbot --nginx --redirect --no-eff-email --email "$email" -d "$PMA_FQDN" || FAILED=true
+  fi
+
   # Check if it succeded
-  if [ ! -d "/etc/letsencrypt/live/$FQDN/" ] || [ "$FAILED" == true ]; then
+  if [ ! -d "/etc/letsencrypt/live/$FQDN/" ] && [ ! -d "/etc/letsencrypt/live/$PMA_FQDN/" ] || [ "$FAILED" == true ]; then
```
Linux123123, line 821: That will not work as it would fail, if the user isn't installing PMA and using letsencrypt.

Linux123123 (changes requested):

vilhelmprytz:
This PR now has conflicts with `master`, do you think you could resolve the conflicts?

Linux123123:
This is very outdated. If any work continues, please open a new PR.

PR #286, opened 2022-01-05
panel: Add listen [::]:80 and listen [::]:443 to nginx configs

### LINKED ISSUE #285: panel: nginx configs does not use IPv6
**Describe the bug**
Nginx config should by default listen to IPv6 addresses as well.

**To Reproduce**

1. Install using installation script.
2. Try to access the site over IPv6 (if the server has it), Nginx does not listen to 80/443 on IPv6 and will therefore not work.

**Expected behaviour**
Script to, by default, listen to IPv6 even though IPv6 may not be available.

**Screenshots**
N/A

 - OS: all affected
 - Installation: panel

**Additional context**
N/A

vilhelmprytz:
fixes #285

### DIFF
--- a/configs/nginx.conf
+++ b/configs/nginx.conf
@@ -1,5 +1,7 @@
 server {
     listen 80;
+    listen [::]:80;
+
     server_name <domain>;
 
     root /var/www/pterodactyl/public;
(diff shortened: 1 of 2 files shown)


vilhelmprytz:
Tested to work, merging.

PR #300, opened 2022-03-05
[panel] Check if FQDN is an IP

Linux123123:
Add's a regex check similar to email, to check if provided FQDN is an IP.

### DIFF
--- a/install-panel.sh
+++ b/install-panel.sh
@@ -67,6 +67,7 @@ user_lastname=""
 user_password=""
 
 # Assume SSL, will fetch different config if true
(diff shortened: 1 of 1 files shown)


Linux123123:
This is only panel, because it would require quite a bit more work on `wings` script.

vilhelmprytz (commented):
I just tried this locally and it still allows me to proceed with automatic Let's Encrypt. Without looking too much at the code, have you verified that this is working?

Linux123123:
I just retested and that regex doesn't work that well. So I just added a much simpler ip checking with `ip`.

PR #315, opened 2022-03-25
Add lib script, rocky and almalinux support

Linux123123:
WIP, not sure how #280 PR will adapt  / be merged.

Currently `panel.sh` has been tested in Ubuntu 20.04 and is at least partially working.

I moved a lot of redundant logic from scripts to lib, so I may have forgotten about something. My goal is to unify as much as possible all the scripts to have the same functions available to minimize overlap.

- [x] `panel`
- [x] `wings`

PR #344, opened 2022-07-13
WIP: Major refactor (split UI and lib, Rocky Linux and AlmaLinux support)

### LINKED ISSUE #283: Rocky Linux
It would be nice to have rocky. linux support added since its basically centos 8. Would it be possible to do this?

vilhelmprytz:
This is still WIP.

Reference #315, thanks @Linux123123 

Fixes #283


install.sh
```diff
@@ -122,3 +95,6 @@ while [ "$done" == false ]; do
   [[ ! " ${valid_input[*]} " =~ ${action} ]] && error "Invalid option"
   [[ " ${valid_input[*]} " =~ ${action} ]] && done=true && IFS=";" read -r i1 i2 <<<"${actions[$action]}" && execute "$i1" "$i2"
 done
+
+# Remove lib.sh, so next time the script is run the newest version is downloaded.
+rm -rf /tmp/lib.sh
```
vilhelmprytz, line 100: If the script fails and exists before this, the file won't be "re-downloaded". I guess that's a very niche case and probably not going to cause problems, but worth considering.

Linux123123: I don't have any other solutions coming up at the moment. If the script fails, it shouldn't be run again, as it's then a not clean system and would fail in other parts. But yeah, a very niche problem.

Linux123123:
<table>

<table>
<tr>
	<td> Panel
	<td> Wings
<tr>
	<td> Ubuntu 22.04, with Let's encrypt, fully works.
	<td>
<tr>
	<td> Rocky Linux 9, with Let's encrypt, fully works.
	<td>
</table>

</table>

Linux123123:
Ready for review. I have tested most of the changes, updated the README, and checked as much of the code as possible for any obvious errors. As long as I am concerned, it is ready to be released.

Taking about releasing, this will change `install.sh` and so change the file paths. Using a prior release (prior to when this is going to be released) will result in an error.

Overall this PR, should improve the speed of the script (as all the packages are installed at once and updating the mirrors is only done when needed), improve the development experience as most things are now included in a lib script and make adding new OS's easier. Also, it is the framework for `phpmyadmin` and `feature install`.

Linux123123 (approved):
LGTM

PR #345, opened 2022-07-18
update uninstall.sh, compatibility to ubuntu 22.04

WirtaxYT:
compatibility to ubuntu 22.04, I have checked the compatibility by modifying the raw and it has worked perfectly in ubuntu 22.04

compatibility to ubuntu 22.04

Fixes #??
compatibility to ubuntu 22.04

### DIFF
--- a/uninstall.sh
+++ b/uninstall.sh
@@ -138,7 +138,7 @@ check_os_comp() {
   case "$OS" in
   ubuntu)
     [ "$OS_VER_MAJOR" == "18" ] && SUPPORTED=true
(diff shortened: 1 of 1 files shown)


vilhelmprytz (changes requested):
Thanks for the contribution! We'd like to support both Ubuntu 20.04 and 22.04, but this PR allows 22.04 while removing 20.04. Do you think you could modify the PR to allow both?

WirtaxYT:
Well, it is not necessary, the already modified command works without errors, I have tested it myself

Linux123123:
LGTM

PR #430, opened 2023-08-30
Added support for Ubuntu 23.04

nickdevnl:
Added support for Ubuntu 23.04


vilhelmprytz:
We intentionally only support LTS releases of Ubuntu, since that is what most people go for. Is there any reason you think 23.04 should be added to the list? It's not an LTS release.

nickdevnl:
Nah, It's just good working no issues with it so far.

Linux123123:
I see only more work and hassle in this. I would not approve this personally.

PR #477, opened 2024-05-09
ubuntu 24.04 support

thewanderingtraderm:
support for ubuntu 24.04 support

Fixes #??


Linux123123:
Don't change the GITHUB_BASE_URL to your fork and from what it looks like, Ubuntu 24,04 has a newer PHP version then pterodactyl supports, so an older version needs to be added from PPA.

Linux123123:
Also workflow needs to be updated (even though it's broken currently), vagrant file too.

PR #482, opened 2024-06-01
github actions: change runs-on from macos -> ubuntu

vilhelmprytz:


Linux123123:
How about we use “docker” driver for all of our images and run everything under docker? In docker there would be all images and I don't see any downsides. We don't actually run wings in the test script, so that wouldn't cause an issue.

vilhelmprytz:
Hmm, so we would then not use full virtualisation at all? Or do you mean trying to run Vagrant inside of Docker?

Linux123123:
Is there any difference if we have fully virtualized, or we have containerized OS? Actions just don't have proper support for Virtualbox. And no vagrant wouldn't work any better in a docker container in an Action.

PR #507, opened 2024-08-26
Update panel.sh

Hamrusy:
Fixes #??


sinjs:
Why?

Linux123123:
There is no reason to change that and at that place, it shouldn't install anything, only configure repositories.

PR #529, opened 2025-04-04
Europe/Kiev renamed to Europe/Kyiv

Waenara:
This pull request fixes an error during the installation process by renaming Europe/Kiev renamed to Europe/Kyiv as Europe/Kiev is not a valid PHP timezone.


Linux123123:
Thanks for your contribution. I have checked the PHP docs, and you are correct.

PR #533, opened 2025-05-25
Add PHPMyAdmin install option

### LINKED ISSUE #194: Add a phpmyadmin install option to the script
**Description**
It would be interesting to add an option to install phpmyadmin to this script.
Something like:
10 - Install PhpMyAdmin. (Use only when the panel has already been installed)

**Why add?**
I think that would be a good thing, because PhpMyAdmin is such a powerful database management tool, which in turn, there are a lot of features that people can use that the dashboard by default doesn't have.

ressidell:
Fixes #194 

### Added

- installers/phpmyadmin.sh, ui/phpmyadmin.sh
  - New ui and installer script to handle phpmyadmin install, I used the already existing panel scripts as a broad template
- PHP-FPM socket file for the CentOS derivatives: configs/www-phpmyadmin.conf

### Changed

- configs/nginx(_ssl).conf
  - in the ssl conf the cert folder was set to the let's encrypt default
  - made the website root and the log name a placeholder, this way the conf can be reused for PMA
- installers/panel.sh
  - placeholder handling from the previous item
- install.sh
  - additional menu items for the PMA installer
  - modified the "execute" function with an additional string split, it was failing with 3 actions
- lib/lib.sh
  - PMA download link and latest release version gathering
    - I couldn't find a nice way to scrape the latest version from the PMA site, but with their Github repo the "get_latest_release" function can be reused

### DIFF
--- a/ui/phpmyadmin.sh
+++ b/ui/phpmyadmin.sh
@@ -135,7 +135,7 @@ main() {
   echo -e -n "\n* Initial configuration completed. Continue with installation? (y/N): "
   read -r CONFIRM
   if [[ "$CONFIRM" =~ [Yy] ]]; then
-    run_installer "panel"
+    run_installer "phpmyadmin"
   else
     error "Installation aborted."
     exit 1


vilhelmprytz:
@Linux123123 Any news on this? :)

